PR #3, opened 2017-03-09
Add german locale

thorbenegberts:
I translated the "assign" with "zuordnen" rather than "zuweisen", which sounds a little too much like a commanding tone. Sadly, in any variant there will be two words for the translation of "unassign" ("Zuordnung aufheben"). This lets the assign button become equally wide as the bookmark button. However, it seems to fit and is not too ugly (see screenshot)

bildschirmfoto 2017-03-09 um 16 43 02

The rest of the texts are as short as possible… Sometimes it's hard to write short sentences in German 😄

### DIFF
--- a/config/locales/client.de.yml
+++ b/config/locales/client.de.yml
@@ -1,4 +1,4 @@
-en:
+de:
   js:
     action_codes:
       assigned: "zugeordnet %{who} %{when}"
--- a/config/locales/client.de.yml
+++ b/config/locales/client.de.yml
@@ -1,0 +1,19 @@
+en:
+  js:
+    action_codes:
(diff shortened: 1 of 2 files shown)


config/locales/client.de.yml
```diff
@@ -0,0 +1,19 @@
+en:
```
ZogStriP: This should be `de`

thorbenegberts:
Sorry, fixed it. Just saw that Sam also corrected in [in my other PR](https://github.com/discourse/discourse-voting/commit/42ffd99c5fc110342fd847c15131d5a1798bd391).

SamSaffron:
np, looks good to me

PR #9, opened 2017-11-13
Added email notification on topic assignment

spog163:
Hi, could you add the email notification on topic asssignment to your amazing plugin, please?
Greetings from Russia:)


coding-horror:
This definitely would need to be a config option default off.

SamSaffron:
@spog163 let us know once this is setup as a site setting.

lib/topic_assigner.rb
```diff
@@ -110,6 +112,9 @@ def assign(assign_to, silent: false)
       user_ids: staff_ids
     )
 
+    message = AssignMailer.send_assignment(assign_to.email, @topic, @assigned_by)
```
tgxworld: Should this unconditionally send out the email even if the user mutes the topic?

rapekas: Let me answer, @spog163 and I worked on this improvement. Currently it sends emails always without an option to disable. Our task was pointed to staff users, who need to have an email notification always. Agree, that this setting must be an option (disabled by default) and depend on mute conditions. So, let us to make some updates first :)

spog163:
@SamSaffron, @coding-horror Added global options as site settings + russian translation (@rapekas)

config/locales/server.en.yml
```diff
@@ -10,8 +10,21 @@ en:
     unassign_on_group_archive: "When a message is archived by a group, unassign message (reassign if moved back to inbox)"
     unassign_on_close: "When a topic is closed unassign topic"
     assign_locks_flags: "When a topic is assigned to a staff member, its flags can only be handled by that person"
+    assign_mailer_enabled: "When enabled, the assigned user will receive a notification email on each assignment"
+    assign_mailer_disabled_for_muted_topics: "When enabled, the assigned user will not receive emails on muted topics"
```
SamSaffron: This is overdoing it I would remove this option, if email_on_assignment is enabled then it should never assign on muted topics you should not be allowed to override it.

rapekas: Thanks for reviewing! Actually user always receive emails on muted topics if 'assign_mailer_disabled_for_muted_topics' disabled or absent. So it is not overrided by any other site setting.

SamSaffron: Sure I understand, I just feel it is an option we don't need to provide.... I would like to delete the setting "assign_mailer_disabled_for_muted_topics"

rapekas:
@SamSaffron Ok, we removed the override option for muted topics, please review.

ZogStriP:
Looks great 👍 It's missing a test though.

SamSaffron:
Thanks @rapekas maybe if you have some time a bit later submit a PR for the test, thanks for the contribution!

PR #14, opened 2018-05-02
FEATURE: Auto assign first staff member

joseanor:


plugin.rb
```diff
@@ -17,6 +17,16 @@
 after_initialize do
   require 'topic_assigner'
 
+  # Assign the first staff member who replies to the topic as the owner.
+  DiscourseEvent.on(:post_created) do |post|
+    topic = Topic.find_by(id: post.topic_id)
+    poster = User.find_by(id: post.user_id)
+    if SiteSetting.assign_first_staff? && topic.custom_fields['assigned_to_id'].nil? && poster.staff?
+      assigner = TopicAssigner.new(topic, poster)
```
ZogStriP: You can do that on one line

```ruby
TopicAssigner.new(topic, poster).assign(poster)
```

plugin.rb
```diff
@@ -17,6 +17,16 @@
 after_initialize do
   require 'topic_assigner'
 
+  # Assign the first staff member who replies to the topic as the owner.
+  DiscourseEvent.on(:post_created) do |post|
+    topic = Topic.find_by(id: post.topic_id)
```
ZogStriP: `post.topic` and `post.user` should return the same thing

ZogStriP:
Thanks 👍 Can you also add test(s)?

plugin.rb
jose-hms: Ah that makes sense.

jose-hms:
Will do!

joseanor:
Closing this PR. I don't have time to add the pertinent test and there are other issues with the execution.

PR #15, opened 2018-11-21
FIX: Unassign on autoclose.

nbianca:


plugin.rb
```diff
@@ -209,8 +209,8 @@ class ::ListController
     ::TopicAssigner.auto_assign(post, force: true)
   end
 
-  on(:topic_closed) do |topic|
-    if SiteSetting.unassign_on_close
+  on(:topic_status_updated) do |topic, status|
+    if SiteSetting.unassign_on_close && (status == 'closed' || status == 'autoclosed')
```
ZogStriP: Can also be written as ;)

```suggestion
    if SiteSetting.unassign_on_close && status["closed"]
```

spec/lib/topic_assigner_spec.rb
```diff
@@ -82,4 +82,28 @@ def assert_publish_topic_state(topic, user)
     end
 
   end
+
+  context "unassign_on_close" do
```
ZogStriP: 😍

PR #16, opened 2018-11-21
FIX: Schedule a rerender on message arrival.

nbianca:


assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
```diff
@@ -153,42 +153,24 @@ function initialize(api) {
     }
   });
 
-  api.createWidget("assigned-to", {
```
ZogStriP: Wow, why was this widget declared 3 times? 😲

assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
```diff
@@ -153,42 +153,24 @@ function initialize(api) {
...
-      ]);
-    }
-  });
-
   api.decorateWidget("post-contents:after-cooked", dec => {
     if (dec.attrs.post_number === 1) {
       const postModel = dec.getModel();
       if (postModel) {
+        // Redraw widget when a message arrives.
+        const messageBus = container.lookup("message-bus:main");
```
ZogStriP: Is there a way we can only subscribe to this channel only once (instead of 3 times)?

assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
SamSaffron: ooooops

assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
SamSaffron: yeah this is a very good catch @ZogStriP  flagged topic listener UX will never display with topic UX as far as I can see so the "blast" unsubscribe everything from the channel will not cause side effects, the blast unsubscribe here will cause issues. I am merging in and fixing.

PR #19, opened 2018-11-27
FIX: assign state was not correctly updated on mobile

jjaffeux:


jjaffeux:
@SamSaffron this commit needs https://github.com/discourse/discourse/pull/6681 and https://github.com/discourse/discourse/pull/6682

PR #20, opened 2018-12-03
FIX: don't assign system user when automatic assign by mention is enabled

majakomel:


lib/topic_assigner.rb
```diff
@@ -115,6 +115,7 @@ def self.mentioned_staff(post)
     mentions = post.raw_mentions
     if mentions.present?
       User.where('moderator OR admin')
+        .where('id > 0')
```
tgxworld:
```suggestion
        .human_users
```

ZogStriP:
Nice 👍

PR #21, opened 2018-12-06
do not assign to same user again

mrfinch:
Also, how to run the unit tests for this engine?

I tried by cloning this code in discourse and running specs there. Is there a better way to do it?
Shouldn't tests be run on PR in this repo too?

### DIFF
--- a/spec/requests/assign_controller_spec.rb
+++ b/spec/requests/assign_controller_spec.rb
@@ -1,9 +1,9 @@
 require 'rails_helper'
 
-RSpec.describe AssignController do
+RSpec.describe DiscourseAssign::AssignController do
 
-  let(:user) { Fabricate(:active_user) }
-  let(:topic) { Fabricate(:topic) }
+  let(:user) { Fabricate(:admin) }
+  let(:post) { Fabricate(:post) }
   let(:user2) { Fabricate(:active_user) }
 
   context 'assign' do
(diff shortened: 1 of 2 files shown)


spec/requests/assign_controller_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  context 'assign' do
+
+    it 'assigns topic to a user' do
+      sign_in(user)
+
+      put '/assign/assign', params: {
+        topic_id: post.topic_id, username: user2.username
+      }
+
+      expect(response.status).to eq(200)
```
SamSaffron: minor but worth a followup PR I would recommend testing that the topic actually got assigned, if we are already doing all the work of assigning the topic.

SamSaffron:
Thanks! looks good.

spec/requests/assign_controller_spec.rb
mrfinch: hey, is there any better way to run unit-tests for this directly here? 

Also, I wanted my message to be i18n supported? Should I first create an entry in discourse in en.yml file and then add a corresponding entry here?

SamSaffron: do you mind heading to https://review.discourse.org/t/fix-do-not-assign-to-same-when-already-assigned/574 I will put the feedback there ... I want to mark it for followup anyway

PR #24, opened 2019-01-08
FEATURE: Add events for assigning and unassigning topics.

nbianca:


SamSaffron:
This is good, can you do a followup PR to add a tiny test to ensure this remains stable over time.

PR #25, opened 2019-01-28
FIX: Check if topic is assigned before assigning again.

nbianca:


SamSaffron:
not sure about this, the instructions should come from discourse code review, it should either: 

1. check if topic is already assigned then not raise the event
OR
2. we add another optional param to the event which allows you only to assign if it is unassigned

PR #26, opened 2019-02-07
FEATURE: uses new topic footer button API

jjaffeux:


jjaffeux:
Conflicts should be good @ZogStriP 

@romanrizzi can you review this and merge it if it looks good, just want to make sure it's not colliding with any of your recent work, thanks.

PR #27, opened 2019-04-18
FEATURE: Display PMs in assigned activity and differentiate them.

romanrizzi:
Screen Shot 2019-04-18 at 15 28 00


assets/javascripts/discourse-assign/controllers/user-activity-assigned.js.es6
```diff
@@ -21,10 +23,25 @@ export default Ember.Controller.extend({
...
               type: "PUT",
               data: { user_id: user.get("id") }
-            }).then(() => this.send("unassignedAll"));
+            }).then(() => this.send("changeAssigned"));
           }
         }
       );
+    },
+    unassign(topic) {
+      ajax("/assign/unassign", {
```
SamSaffron: hmmm isn't this ajax call elsewhere, should we not be importing a model of sorts and just calling it?

assets/javascripts/discourse/templates/components/assigned-topic-list-item.hbs
```diff
@@ -0,0 +1,53 @@
+{{#if bulkSelectEnabled}}
+  <td class="bulk-select">
```
SamSaffron: do we need bulk select here?

assets/javascripts/discourse/components/assigned-topic-list-item.js.es6
```diff
@@ -0,0 +1,187 @@
...
+  },
+
+  @computed("topic.archetype")
+  isPrivateMessage(archetype) {
+    return archetype === "private_message";
+  },
+
+  actions: {
+    toggleBookmark() {
+      this.get("topic").toggleBookmark();
```
SamSaffron: do we need this in the list?

assets/javascripts/discourse/components/assigned-topic-list-item.js.es6
```diff
@@ -0,0 +1,187 @@
...
+  }
+}
+
+export function navigateToTopic(topic, href) {
+  this.appEvents.trigger("header:update-topic", topic);
+  DiscourseURL.routeTo(href || topic.get("url"));
+  return false;
+}
+
+export default Ember.Component.extend({
```
SamSaffron: Can we not just inherit off core here and reuse the majority of the implementation?

SamSaffron:
My overriding concern here is that we are cutting and pasting a lot of code from core, I wonder if there are better ways of reusing bits, @eviltrout any ideas here?

eviltrout:
Yes when @romanrizzi mentioned this on our call I thought it was just the template being copied over, but here I see there's a lot of Javascript code too.

Could we not use the component that renders the topic list and use a plugin outlet for the assigned specific controls? Or at least extend it like sam mentions?

romanrizzi:
I came up with a different approach here:

https://github.com/discourse/discourse/pull/7433

That PR aims to decouple the default behavior of a `topic-list-item` from the component itself, making it easy to re-use it from another component defining a non-raw version of that component.

This way we can rewrite the `assign-topic-list-item` component as: 

```javascript
Ember.Component.extend(ListItemDefaults)
```

SamSaffron:
@eviltrout when you get back can you take care of merging this + the core PR ?

tgxworld:
Does this replace the filter I added for messages?

Screenshot from 2019-04-29 16-02-50

romanrizzi:
@tgxworld The messages section uses a different route and will remain unchanged.

If we merge the code PR, we need to be aware that it won't be available until `beta10`. What's the best strategy to avoid breaking instances running `beta9` and `discourse-assign`?

eviltrout:
@romanrizzi I've approved the core PR so I presume this now needs to be updated?

romanrizzi:
@eviltrout Done!

romanrizzi:
I was thinking about this yesterday.

I don't know if this is possible in our codebase but I think we can check if the Mixin is present using the `require` primitive and then decide what we're gonna do.

We can ship a minimal `ListItemDefault` object and fallback to it if necessary (and raise a warning).

 Versioning this plugin would be awesome to avoid shipping breaking changes because we depend on core's code.

eviltrout:
@romanrizzi I do believe in the past we looked at pinning plugins to branches. I would look into that and see if we are still doing it (for example stable / beta.) If not the require solution might work. You could also consider comitting an empty version of the mixin to the stable/beta branches so the import does not fail, but check if the object is empty.

Feel free to discuss on dev / meta if you can find a topic about plugin/version maintenance.

SamSaffron:
Thanks Roman, trying this out!

PR #28, opened 2019-04-22
FEATURE: Remind users of assigned tasks.

romanrizzi:
### Changes

- New `SiteSetting` to set reminders frequency.
- Sends a PM as a reminder of assigned tasks:
  - Only if `tasks >= 1`.

Screen Shot 2019-04-17 at 10 10 05
Screen Shot 2019-04-17 at 10 22 33

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -12,7 +12,6 @@ en:
       unassign_all:
         title: "Unassign All"
         confirm: "Are you sure you want to unassign all topics from {{username}}?"
-
       unassign:
         title: "Unassign"
         help: "Unassign Topic"
@@ -26,6 +25,11 @@ en:
       claim:
         title: "claim"
         help: "Assign topic to yourself"
+      reminders_frequency:
+        never: 'Never'
+        daily: 'Daily'
+        monthly: 'Monthly'
+        quarterly: 'Quarterly'
     user:
       messages:
         assigned_title: "Assigned (%{count})"
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -12,6 +12,7 @@ en:
     assign_locks_flags: "When a topic is assigned to a staff member, its flags can only be handled by that person"
     assign_mailer_enabled: "When enabled, the assigned user will receive a notification email on each assignment"
     flags_require_assign: "When enabled, flags cannot be handled unless assigned to a user."
+    remind_assigns: "Remind users about pending assigns."
   discourse_assign:
     assigned_to: "Topic assigned to @%{username}"
     unassigned: "Topic was unassigned"
@@ -31,3 +32,22 @@ en:
 
       If you're interested, click the link below:
       [%{topic_link}](%{topic_link})
+  pending_assigns_reminder:
+    title: "You have %{pending_assignments} pending assignments"
+    body: |
+      You currently have [%{pending_assignments} pending assignments](%{assignments_link}).
+
+      %{newest_assignments}
+      %{oldest_assignments}
+      
+      This reminder will be sent %{frequency} if you have more than one assigned topic.
+    newest: |
+      ### Newest
+      %{topic_0}
+      %{topic_1}
+      %{topic_2}
+    oldest: |
+      ### Oldest
+      %{topic_0}
+      %{topic_1}
+      %{topic_2}
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -13,6 +13,7 @@ en:
     assign_mailer_enabled: "When enabled, the assigned user will receive a notification email on each assignment"
     flags_require_assign: "When enabled, flags cannot be handled unless assigned to a user."
     remind_assigns: "Remind users about pending assigns."
+    remind_assigns_frequency: "Frequency for reminding users about assigned topics."
   discourse_assign:
     assigned_to: "Topic assigned to @%{username}"
     unassigned: "Topic was unassigned"
@@ -21,6 +22,11 @@ en:
     flag_assigned: "Sorry, that flag's topic is assigned to another user"
     flag_unclaimed: "You must claim that topic before acting on the flag"
     topic_assigned_excerpt: "assigned you the topic '%{title}'"
+    reminders_frequency:
+        never: 'never'
+        daily: 'daily'
+        monthly: 'monthly'
+        quarterly: 'quarterly'
   assign_mailer:
     title: "Assign Mailer"
     subject_template: "[%{email_prefix}] %{assignee_name} assigned you to '%{topic_title}'!"
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -19,3 +19,13 @@ plugins:
     default: false
     client: true
   assign_mailer_enabled: false
+  remind_assigns:
+    client: true
+    default: 'monthly'
+    type: enum
+    choices:
+      - 'never'
+      - 'daily'
+      - 'weekly'
+      - 'monthly'
+      - 'quarterly'
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -19,13 +19,7 @@ plugins:
     default: false
     client: true
   assign_mailer_enabled: false
-  remind_assigns:
+  remind_assigns_frequency:
     client: true
-    default: 'monthly'
-    type: enum
-    choices:
-      - 'never'
-      - 'daily'
-      - 'weekly'
-      - 'monthly'
-      - 'quarterly'
+    enum: "RemindAssignsFrequencySiteSettings"
+    default: 43200
--- a/db/migrate/20190422200243_add_last_reminded_at_index.rb
+++ b/db/migrate/20190422200243_add_last_reminded_at_index.rb
@@ -1,0 +1,13 @@
+class AddLastRemindedAtIndex < ActiveRecord::Migration[5.2]
+  def up
+    execute <<~SQL
+    CREATE UNIQUE INDEX idx_last_reminded_at
+    ON user_custom_fields(name, user_id)
+    WHERE name = 'last_reminded_at'
+    SQL
+  end
+
+  def down
+    execute "DROP INDEX idx_last_reminded_at"
+  end
+end
--- a/db/migrate/20190422200243_add_last_reminded_at_index.rb
+++ b/db/migrate/20190422200243_add_last_reminded_at_index.rb
@@ -1,13 +1,6 @@
 class AddLastRemindedAtIndex < ActiveRecord::Migration[5.2]
-  def up
-    execute <<~SQL
-    CREATE UNIQUE INDEX idx_last_reminded_at
-    ON user_custom_fields(name, user_id)
-    WHERE name = 'last_reminded_at'
-    SQL
-  end
-
-  def down
-    execute "DROP INDEX idx_last_reminded_at"
+  def change
+    add_index :user_custom_fields, %i[name user_id], name: :idx_user_custom_fields_last_reminded_at,
+                                                     unique: true, where: "name = 'last_reminded_at'"
   end
 end
--- a/db/migrate/20190422200243_add_last_reminded_to_user.rb
+++ b/db/migrate/20190422200243_add_last_reminded_to_user.rb
@@ -1,5 +1,0 @@
-class AddLastRemindedToUser < ActiveRecord::Migration[5.2]
-  def change
-    add_column :users, :last_tasks_reminder, :datetime
-  end
-end
--- a/db/migrate/20190422200243_add_last_reminded_to_user.rb
+++ b/db/migrate/20190422200243_add_last_reminded_to_user.rb
@@ -1,0 +1,5 @@
+class AddLastRemindedToUser < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :last_tasks_reminder, :datetime
+  end
+end
--- a/jobs/regular/remind_user.rb
+++ b/jobs/regular/remind_user.rb
@@ -1,0 +1,13 @@
+module Jobs
+  class RemindUser < Jobs::Scheduled
+    sidekiq_options queue: 'low'
+
+    def execute(args)
+      raise Discourse::InvalidParameters.new(:user_id) unless args[:user_id].present?
+
+      user = User.find(args[:user_id])
+
+      PendingAssignsReminder.new.remind(user)
+    end
+  end
+end
--- a/jobs/regular/remind_user.rb
+++ b/jobs/regular/remind_user.rb
@@ -1,11 +1,10 @@
 module Jobs
-  class RemindUser < Jobs::Scheduled
+  class RemindUser < Jobs::Base
     sidekiq_options queue: 'low'
 
     def execute(args)
-      raise Discourse::InvalidParameters.new(:user_id) unless args[:user_id].present?
-
-      user = User.find(args[:user_id])
+      user = User.find_by(args[:user_id])
+      raise Discourse::InvalidParameters.new(:user_id) if user.nil?
 
       PendingAssignsReminder.new.remind(user)
     end
--- a/jobs/scheduled/enqueue_reminders.rb
+++ b/jobs/scheduled/enqueue_reminders.rb
@@ -1,5 +1,7 @@
 module Jobs
   class EnqueueReminders < Jobs::Scheduled
+    every 2.hours
+
     def execute(_args)
       return if skip_enqueue?
       user_ids.each { |id| Jobs.enqueue(:remind_user, user_id: id) }
@@ -8,11 +10,11 @@ module Jobs
     private
 
     def skip_enqueue?
-      SiteSetting.remind_assigns.nil? || SiteSetting.remind_assigns == 'never'
+      SiteSetting.remind_assigns_frequency.nil? || SiteSetting.remind_assigns_frequency.zero?
     end
 
     def user_ids
-      interval = reminder_interval_in_minutes(SiteSetting.remind_assigns)
+      interval = SiteSetting.remind_assigns_frequency
 
       TopicCustomField
         .joins(<<~SQL
@@ -21,24 +23,12 @@ module Jobs
         SQL
         ).where(<<~SQL
           user_custom_fields.value IS NULL OR
-          COALESCE(user_custom_fields.value, '2010-01-01')::TIMESTAMP <= CURRENT_TIMESTAMP - ('1 MINUTE'::INTERVAL * #{interval})
+          user_custom_fields.value::TIMESTAMP <= CURRENT_TIMESTAMP - ('1 MINUTE'::INTERVAL * #{interval})
         SQL
-        ).where(name: TopicAssigner::ASSIGNED_TO_ID)
+        ).where("topic_custom_fields.updated_at::TIMESTAMP <= CURRENT_TIMESTAMP - ('1 MINUTE'::INTERVAL * ?)", interval)
+        .where(name: TopicAssigner::ASSIGNED_TO_ID)
         .group('topic_custom_fields.value').having('COUNT(topic_custom_fields.value) > 1')
         .pluck('topic_custom_fields.value')
-    end
-
-    def reminder_interval_in_minutes(remind_frequency)
-      case remind_frequency
-      when 'daily'
-        1440
-      when 'weekly'
-        10080
-      when 'monthly'
-        43200
-      else
-        131400 # quarterly
-      end
     end
   end
 end
--- a/jobs/scheduled/enqueue_reminders.rb
+++ b/jobs/scheduled/enqueue_reminders.rb
@@ -1,0 +1,38 @@
+module Jobs
+  class EnqueueReminders < Jobs::Scheduled
+    def execute(_args)
+      return if skip_enqueue?
+      user_ids.each { |id| Jobs.enqueue(:remind_user, user_id: id) }
+    end
+
+    private
+
+    def skip_enqueue?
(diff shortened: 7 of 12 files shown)


db/migrate/20190422200243_add_last_reminded_to_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddLastRemindedToUser < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :last_tasks_reminder, :datetime
```
SamSaffron: We can not be bolting on columns to the user table 

Use a user custom field here and add a unique index on it.

config/locales/server.en.yml
```diff
@@ -31,3 +32,22 @@ en:
 
       If you're interested, click the link below:
       [%{topic_link}](%{topic_link})
+  pending_assigns_reminder:
+    title: "You have %{pending_assignments} pending assignments"
+    body: |
+      You currently have [%{pending_assignments} pending assignments](%{assignments_link}).
+
+      %{newest_assignments}
+      %{oldest_assignments}
```
SamSaffron: I would only have N oldest with a (7 more) link at the bottom.  Keep it to say 10 + an optional N more link.

jobs/scheduled/enqueue_reminders.rb
```diff
@@ -0,0 +1,38 @@
...
+    def skip_enqueue?
+      SiteSetting.remind_assigns.nil? || SiteSetting.remind_assigns == 'never'
+    end
+
+    def user_ids
+      interval = reminder_interval_in_minutes(SiteSetting.remind_assigns)
+
+      TopicCustomField
+        .joins("LEFT JOIN users ON topic_custom_fields.value::INT = users.id")
+        .where("COALESCE(users.last_tasks_reminder, '2010-01-01') <= CURRENT_TIMESTAMP - ('1 MINUTE'::INTERVAL * ?)", interval)
```
SamSaffron: I kind of feel bummed to be reminding people of work that was just assigned to them today. 

I would add one more where clause here to ensure the thing was assigned prior to current_date - interval.

lib/pending_assigns_reminder.rb
```diff
@@ -0,0 +1,59 @@
...
+      'pending_assigns_reminder.body',
+      pending_assignments: topics.size,
+      assignments_link: "#{Discourse.base_url}/u/#{user.username_lower}/assigned",
+      newest_assignments: newest_list,
+      oldest_assignments: oldest_list,
+      frequency: SiteSetting.remind_assigns
+    )
+  end
+
+  def build_list_for(key, topics)
```
SamSaffron: Can core provide better facilities here?

romanrizzi: Couldn't find any helper in core. Am I missing something?

config/locales/server.en.yml
romanrizzi: Could you please explain? I'm not sure I'm understanding your point.

db/migrate/20190422200243_add_last_reminded_at_index.rb
```diff
@@ -0,0 +1,13 @@
+class AddLastRemindedAtIndex < ActiveRecord::Migration[5.2]
+  def up
+    execute <<~SQL
+    CREATE UNIQUE INDEX idx_last_reminded_at
```
tgxworld: Usually we include the table name in the index name so that we can quickly tell which table it belongs to.

jobs/regular/remind_user.rb
```diff
@@ -0,0 +1,13 @@
+module Jobs
+  class RemindUser < Jobs::Scheduled
+    sidekiq_options queue: 'low'
+
+    def execute(args)
+      raise Discourse::InvalidParameters.new(:user_id) unless args[:user_id].present?
+
+      user = User.find(args[:user_id])
```
tgxworld: You'll want to use `find_by` here instead and raise a `Discourse::InvalidParameters` error if the `user_id` is no longer valid.

config/settings.yml
```diff
@@ -19,3 +19,13 @@ plugins:
     default: false
     client: true
   assign_mailer_enabled: false
+  remind_assigns:
+    client: true
+    default: 'monthly'
+    type: enum
+    choices:
```
tgxworld: I think we should extract the choices into a class under `EnumSiteSetitng` so that the values can be accessed without having to duplicate the choices in multiple places.

spec/jobs/scheduled/enqueue_reminders_spec.rb
```diff
@@ -0,0 +1,72 @@
+require 'rails_helper'
+
+RSpec.describe Jobs::EnqueueReminders do
+  let(:user) { Fabricate(:user) }
+
+  describe '#execute' do
+    it 'Do not enqueue reminders when there are no assigned tasks' do
+      Jobs.expects(:enqueue).never
```
tgxworld: We don't encourage the use of stubs in our test suite. Sidekiq provides a bunch of useful helpers that you can use. https://github.com/mperham/sidekiq/wiki/Testing

jobs/scheduled/enqueue_reminders.rb
```diff
@@ -0,0 +1,44 @@
+module Jobs
+  class EnqueueReminders < Jobs::Scheduled
```
tgxworld: Do we need a duration for this scheduled job?

jobs/regular/remind_user.rb
```diff
@@ -0,0 +1,13 @@
+module Jobs
+  class RemindUser < Jobs::Scheduled
```
tgxworld: I'm guessing this shouldn't be a scheduled job?

jobs/scheduled/enqueue_reminders.rb
```diff
@@ -0,0 +1,44 @@
...
+    def skip_enqueue?
+      SiteSetting.remind_assigns.nil? || SiteSetting.remind_assigns == 'never'
+    end
+
+    def user_ids
+      interval = reminder_interval_in_minutes(SiteSetting.remind_assigns)
+
+      TopicCustomField
+        .joins(<<~SQL
+          LEFT OUTER JOIN user_custom_fields ON topic_custom_fields.value::INT = user_custom_fields.user_id
```
tgxworld: Is there a specific name for `topic_custom_fields` to join on?

lib/pending_assigns_reminder.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+class PendingAssignsReminder
+  REMINDED_AT = 'last_reminded_at'
+
+  def remind(user)
+    topics = assigned_topics(user)
+    return if topics.size < 2
```
tgxworld: Let's replace `2` with a constant so that it is clear what 2 represents.

lib/pending_assigns_reminder.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  def assigned_topics(user)
+    Topic.joins(:_custom_fields).select(:slug, :id, :fancy_title, 'topic_custom_fields.created_at AS assigned_at')
+      .where('topic_custom_fields.name = ? AND topic_custom_fields.value = ?', TopicAssigner::ASSIGNED_TO_ID, user.id.to_s)
+      .order('topic_custom_fields.created_at DESC')
+      .limit(6)
+  end
+
+  def reminder_body(user, topics)
+    newest, oldest = topics.each_slice(3).to_a
```
tgxworld: Will the second slice be the oldest topics? I see the query is ordered by `created_at DESC` and has a limit of 6. Does that mean `oldest` is just topics that are not top 3?

spec/jobs/scheduled/enqueue_reminders_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      Jobs.expects(:enqueue).with(:remind_user, user_id: user.id.to_s).never
+
+      trigger_enqueue
+    end
+
+    def trigger_enqueue
+      subject.execute({})
+    end
+
+    def assign_one_task_to(an_user)
```
tgxworld:
```suggestion
    def assign_one_task_to(user)
```

spec/jobs/scheduled/enqueue_reminders_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      post = Fabricate(:post)
+      TopicAssigner.new(post.topic, an_user).assign(an_user)
```
tgxworld:
```suggestion
      TopicAssigner.new(post.topic, user).assign(user)
```

spec/jobs/scheduled/enqueue_reminders_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    end
+
+    def assign_multiple_tasks_to(an_user)
```
tgxworld:
```suggestion
    def assign_multiple_tasks_to(user)
```

spec/jobs/scheduled/enqueue_reminders_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      assign_one_task_to(an_user)
```
tgxworld:
```suggestion
      assign_one_task_to(user)
```

spec/jobs/scheduled/enqueue_reminders_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      assign_one_task_to(an_user)
```
tgxworld:
```suggestion
      assign_one_task_to(user)
```

spec/jobs/scheduled/enqueue_reminders_spec.rb
tgxworld: Instead of having to repeat `assign_one_task_to` twice here, we can make use of the new constant and do a loop. That way, any changes to the threshold of assign reminder will not break the tests.

spec/lib/pending_assigns_reminder_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  it { assert_reminder_not_created }
+
+  it 'Do not create a reminder if the user only has one task' do
+    post = Fabricate(:post)
+    TopicAssigner.new(post.topic, user).assign(user)
+
+    assert_reminder_not_created
+  end
+
+  def assert_reminder_not_created
```
tgxworld: This assertion doesn't feel correct. What if `post_created` event is never called? Perhaps we can just assert that no new post has been created?

spec/lib/pending_assigns_reminder_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+        reminded_at = user.reload.custom_fields[described_class::REMINDED_AT].to_datetime
+
+        expect(reminded_at).to eq_time(expected_last_reminder)
+      end
+    end
+
+    def assert_remind_was_created_correctly(topic, post)
+      expect(topic.user).to eq(system)
+      expect(topic.archetype).to eq(Archetype.private_message)
+      expect(topic.topic_allowed_users.map(&:user_id)).to match_array([system.id, user.id])
```
tgxworld:
```suggestion
      expect(topic.topic_allowed_users.pluck(:user_id)).to contain_exactly(system.id, user.id)
```

lib/pending_assigns_reminder.rb
SamSaffron: Oh what I meant is ... is there any code that can be shuffled in core to make this easier for plugins?

romanrizzi:
Thanks @tgxworld 🙌 

- I was selecting the six newest assigns and dividing them into two groups of three. Now I take the newest three and the oldest three.
- The total assigned topics count was fixed.

lib/pending_assigns_reminder.rb
romanrizzi: I don't believe this will be re-used by another plugin 🤔

config/locales/server.en.yml
SamSaffron: Its ok lets just try this format for now.

SamSaffron:
@tgxworld if you are happy with this can you merge it in?

config/locales/server.en.yml
```diff
@@ -12,6 +12,8 @@ en:
     assign_locks_flags: "When a topic is assigned to a staff member, its flags can only be handled by that person"
     assign_mailer_enabled: "When enabled, the assigned user will receive a notification email on each assignment"
     flags_require_assign: "When enabled, flags cannot be handled unless assigned to a user."
+    remind_assigns: "Remind users about pending assigns."
+    remind_assigns_frequency: "Remind users about assigned topics."
```
tgxworld: The description here doesn't seem to match the key.

config/locales/client.en.yml
```diff
@@ -26,6 +25,11 @@ en:
       claim:
         title: "claim"
         help: "Assign topic to yourself"
+      reminders_frequency:
```
tgxworld: I'm not very sure here but is this being used on the client side?

lib/pending_assigns_reminder.rb
```diff
@@ -0,0 +1,86 @@
...
+      pending_assignments: assigned_topics_count,
+      assignments_link: "#{Discourse.base_url}/u/#{user.username_lower}/assigned",
+      newest_assignments: newest_list,
+      oldest_assignments: oldest_list,
+      frequency: frequency_in_words
+    )
+  end
+
+  def build_list_for(key, topics)
+    return '' if topics.nil? || topics.size.zero?
```
tgxworld:
```suggestion
    return '' if topics.empty?
```

spec/jobs/scheduled/enqueue_reminders_spec.rb
```diff
@@ -0,0 +1,71 @@
+require 'rails_helper'
+
+RSpec.describe Jobs::EnqueueReminders do
+  let(:user) { Fabricate(:user) }
+
+  after do
+    monthly_frequency = 43200
+    SiteSetting.remind_assigns_frequency = monthly_frequency
```
tgxworld: There is no need to reset site settings after each test because we use an in memory site settings provider that is automatically cleared after each test.

spec/lib/pending_assigns_reminder_spec.rb
```diff
@@ -0,0 +1,61 @@
+require 'rails_helper'
+
+RSpec.describe PendingAssignsReminder do
```
tgxworld:
```suggestion
RSpec.describe Jobs::PendingAssignsReminder do
```

spec/lib/pending_assigns_reminder_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+  let(:user) { Fabricate(:user) }
+
+  it 'Do not create a reminder if the user has 0 assigned topics' do
```
tgxworld: This is very minor but I think we should maintain the same convention that the rest of our specs are using. There are two things here I think we should change: 

1. Our existing tests do not capitalize the first letter of the description
1. it 'does not ...' instead of it 'do not...'

spec/lib/pending_assigns_reminder_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      @another_post = Fabricate(:post)
+      TopicAssigner.new(@post.topic, user).assign(user)
+      TopicAssigner.new(@another_post.topic, user).assign(user)
+      @assigned_posts = 2
+    end
+
+    it 'Creates a reminder for a particular user' do
+      subject.remind(user)
+
+      created_post = Post.includes(topic: %i[topic_allowed_users]).last
```
tgxworld: Hmm do we need the eager loading here if we're going to planning to pluck the results from the DB?

spec/lib/pending_assigns_reminder_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+
+      assert_remind_was_created_correctly(created_post.topic, created_post)
+    end
+
+    it 'Sets the timestamp of the last reminder' do
```
tgxworld: I would just combine this test with the one above since the outcome is the same.

spec/lib/pending_assigns_reminder_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      TopicAssigner.new(@another_post.topic, user).assign(user)
+      @assigned_posts = 2
+    end
+
+    it 'Creates a reminder for a particular user' do
+      subject.remind(user)
+
+      created_post = Post.includes(topic: %i[topic_allowed_users]).last
+
+      assert_remind_was_created_correctly(created_post.topic, created_post)
```
tgxworld: Minor but we can just pass in the post here and have the method pick out the topic.

jobs/scheduled/enqueue_reminders.rb
```diff
@@ -0,0 +1,34 @@
...
+      interval = SiteSetting.remind_assigns_frequency
+
+      TopicCustomField
+        .joins(<<~SQL
+          LEFT OUTER JOIN user_custom_fields ON topic_custom_fields.value::INT = user_custom_fields.user_id
+          AND user_custom_fields.name = '#{PendingAssignsReminder::REMINDED_AT}'
+        SQL
+        ).where(<<~SQL
+          user_custom_fields.value IS NULL OR
+          COALESCE(user_custom_fields.value, '2010-01-01')::TIMESTAMP <= CURRENT_TIMESTAMP - ('1 MINUTE'::INTERVAL * #{interval})
```
tgxworld: Do we need the `COALESCE` since we have a `IS NULL` check before?

jobs/scheduled/enqueue_reminders.rb
```diff
@@ -0,0 +1,34 @@
...
+        SQL
+        ).where("COALESCE(topic_custom_fields.updated_at, '2010-01-01')::TIMESTAMP <= CURRENT_TIMESTAMP - ('1 MINUTE'::INTERVAL * ?)", interval)
```
tgxworld: Is `topic_custom_fields.updated_at` ever blank?

lib/pending_assigns_reminder.rb
```diff
@@ -0,0 +1,86 @@
+# frozen_string_literal: true
+
+class PendingAssignsReminder
+  REMINDED_AT = 'last_reminded_at'
+  REMINDER_THRESHOLD = 2
+
+  def remind(user)
+    newest_topics = assigned_topics(:newest, user)
+    return if newest_topics.size < REMINDER_THRESHOLD
+    oldest_topics = assigned_topics(:oldest, user).where.not(id: newest_topics.map(&:id))
```
tgxworld: Personally, I would go with something like `assigned_topics(user, order: :asc).where.not(id: newest_topics.map(&:id))`

`def assigned_topics(user, order: :desc)`

I don't feel like we need `:newest, :oldest` to represent ordering here.

lib/pending_assigns_reminder.rb
```diff
@@ -0,0 +1,86 @@
...
+      validate: false
+    )
+
+    update_last_reminded(user)
+  end
+
+  private
+
+  def assigned_count_for(user)
+    TopicCustomField.where(name: TopicAssigner::ASSIGNED_TO_ID, value: user.id.to_s).count
```
tgxworld:
```suggestion
    TopicCustomField.where(name: TopicAssigner::ASSIGNED_TO_ID, value: user.id).count
```

lib/pending_assigns_reminder.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  def assigned_topics(type, user)
+    order = type == :newest ? 'DESC' : 'ASC'
+    Topic.joins(:_custom_fields).select(:slug, :id, :fancy_title, 'topic_custom_fields.created_at AS assigned_at')
+      .where('topic_custom_fields.name = ? AND topic_custom_fields.value = ?', TopicAssigner::ASSIGNED_TO_ID, user.id.to_s)
```
tgxworld:
```suggestion
      .where('topic_custom_fields.name = ? AND topic_custom_fields.value = ?', TopicAssigner::ASSIGNED_TO_ID, user.id)
```

lib/pending_assigns_reminder.rb
```diff
@@ -0,0 +1,86 @@
...
+      memo["topic_#{index}"] = "- [#{Emoji.gsub_emoji_to_unicode(t.fancy_title)}](#{t.relative_url}) - assigned #{time_in_words_for(t)}"
+      memo
+    end
+
+    I18n.t("pending_assigns_reminder.#{key}", items.symbolize_keys!)
+  end
+
+  def time_in_words_for(topic)
+    FreedomPatches::Rails4.distance_of_time_in_words(
+      Time.now, topic.assigned_at.to_time, false, scope: 'datetime.distance_in_words_verbose'
```
tgxworld:
```suggestion
      Time.zone.now, topic.assigned_at.to_time, false, scope: 'datetime.distance_in_words_verbose'
```

tgxworld:
@romanrizzi I did another round of review and left some comments. Feel free to merge once they have been resolved.

lib/pending_assigns_reminder.rb
romanrizzi: If I apply this suggestion, I'll get a type mismatch error since value must be a string.

config/locales/client.en.yml
romanrizzi: It's being used by `RemindAssignsFrequencySiteSettings#frequency_for`

romanrizzi:
@tgxworld Everything except two comments addressed. If you're OK with it, I'll merge later today.

SamSaffron:
OK, I am merging! thanks @romanrizzi !

PR #30, opened 2019-05-03
FEATURE: Users can override reminders frequency

romanrizzi:
Screen Shot 2019-05-03 at 11 52 01

### Description

Users can now adjust the frequency of the reminders introduced in #28.

The select can be found under `Preferences > Notifications`. The global frequency will be selected if the user never overrode this value before.

It will only be visible if the plugin is enabled.

### Changes 

- Added `weekly` option.
- New user custom field: `remind_assigns_frequency` with a unique index.
- When enqueuing reminders:
  - We skip the `enqueue_reminders` job if the plugin is not enabled.
  - We avoid counting deleted topics when selecting users with +1 assigns.
  - We use the user selected frequency or fallback to the global.


assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
```diff
@@ -214,6 +214,15 @@ function initialize(api) {
     "notification.discourse_assign.assign_notification",
     "user-plus"
   );
+
+  api.modifyClass("controller:preferences/notifications", {
+    actions: {
+      save() {
+        this.get("saveAttrNames").push("custom_fields");
+        this._super();
```
jjaffeux:
```suggestion
        this._super(...arguments);
```

romanrizzi:
@tgxworld Mind taking a look?

assets/javascripts/discourse-assign/connectors/user-preferences-notifications/remind-assigns-frequency.hbs
```diff
@@ -0,0 +1 @@
+{{remind-assigns-frequency user=model}}
```
tgxworld: Minor one but there is a missing blank line here

assets/javascripts/discourse/components/remind-assigns-frequency.js.es6
```diff
@@ -0,0 +1,23 @@
+import computed from "ember-addons/ember-computed-decorators";
+
+export default Ember.Component.extend({
+  @computed("user.reminders_frequency")
+  translatedFrequencies() {
```
tgxworld: Something about the name feels odd to me because the property is actually building the content for the combobox. Translating the frequency is a subset of what this component does.

assets/javascripts/discourse/components/remind-assigns-frequency.js.es6
```diff
@@ -0,0 +1,23 @@
...
+  },
+
+  didInsertElement() {
+    const user_frequency = this.get(
+      "user.custom_fields.remind_assigns_frequency"
+    );
+    if (user_frequency) return;
+
+    const global_frequency = this.get("siteSettings.remind_assigns_frequency");
+    this.set("user.custom_fields.remind_assigns_frequency", global_frequency);
```
tgxworld: Wouldn't this insert custom field records to all users that save their user preferences?

jobs/scheduled/enqueue_reminders.rb
```diff
@@ -12,26 +12,39 @@ def execute(_args)
     private
 
     def skip_enqueue?
-      SiteSetting.remind_assigns_frequency.nil? || SiteSetting.remind_assigns_frequency.zero?
+      SiteSetting.remind_assigns_frequency.nil? || !SiteSetting.assign_enabled?
     end
 
     def user_ids
-      interval = SiteSetting.remind_assigns_frequency
+      global_frequency = SiteSetting.remind_assigns_frequency.to_i
```
tgxworld: Do we need the `.to_i` here? My understanding is that `SiteSetting` will cast the value automatically when it retrieves from the DB.

jobs/scheduled/enqueue_reminders.rb
```diff
@@ -12,26 +12,39 @@ def execute(_args)
...
+      frequency = "COALESCE(user_frequency.value, '#{global_frequency}')::INT"
```
tgxworld: We need to be careful here because `user_frequency.value` is a user input and I don't think we actually validate the values that `user_frequency.value` can be configured to. I believe someone can easily submit any custom strings via a carefully crafted `POST` request.

spec/jobs/scheduled/enqueue_reminders_spec.rb
```diff
@@ -49,7 +50,29 @@
...
+        PendingAssignsReminder::REMINDERS_FREQUENCY => RemindAssignsFrequencySiteSettings::DAILY_MINUTES
+      )
+      user.save_custom_fields
+
+      assign_multiple_tasks_to(user)
+
+      assert_reminders_enqueued(1)
+    end
+
+    it "Doesn't count assigns from deleted topics" do
```
tgxworld:
```suggestion
    it "doesn't count assigns from deleted topics" do
```

jobs/scheduled/enqueue_reminders.rb
romanrizzi: This is a good observation. I added a `before_save` callback to coerce the value field and avoid this threat.

lib/pending_assigns_reminder.rb
```diff
@@ -71,8 +72,14 @@ def time_in_words_for(topic)
     )
   end
 
-  def frequency_in_words
-    ::RemindAssignsFrequencySiteSettings.frequency_for(SiteSetting.remind_assigns_frequency)
+  def frequency_in_words(user)
+    frequency = if user.custom_fields && user.custom_fields.has_key?(REMINDERS_FREQUENCY)
```
tgxworld:
```suggestion
    frequency = if user.custom_fields&.has_key?(REMINDERS_FREQUENCY)
```

jobs/scheduled/enqueue_reminders.rb
tgxworld: The `before_save` callback is a good idea but I think to be fully protected, we can just escape the values here.

tgxworld:
@romanrizzi One more comment otherwise this PR looks good to me.

jobs/scheduled/enqueue_reminders.rb
romanrizzi: I've been doing some research about this and I'm not sure what's the best way of escaping the value of `user_frequency.value` 🤔 .

SamSaffron: Yeah one issue here is that your are injecting "SiteSetting.remind_assigns_frequency" which is not ideal cause there is an exploit here potentially waiting to happen. 

Instead just write one big SQL statement here and use `MiniSql.query_single` as an added bonus it will be far easier to comprehend.

tgxworld: @romanrizzi You can use `ActiveRecord::Base.sanitize` to sanitize inputs in the future.

romanrizzi: Oh, we want to sanitize the `SiteSetting`. I thought that we wanted to sanitize the value stored inside `user_frequency.value` (this is covered by the `before_save`).

romanrizzi: Done. I sanitized the `SiteSetting` and rewrote the query with `MiniSql.query_single` 😄

romanrizzi:
Merging in!

romanrizzi:
I'll merge this one on my Monday morning

PR #31, opened 2019-05-03
Feature: Allow assign on groups

romanrizzi:


romanrizzi:
@tgxworld Can you review? Sorry to ask for review so ofter but at this point, you're too familiar with my PRs on this plugin 😛.

assets/javascripts/discourse/services/task-actions.js.es6
```diff
@@ -1,7 +1,16 @@
...
 import showModal from "discourse/lib/show-modal";
+import { getOwner } from "discourse-common/lib/get-owner";
 
 export default Ember.Service.extend({
+  init() {
+    this._super(...arguments);
+
+    this.allowedGroups = getOwner(this)
+      .lookup("site-settings:main")
+      .assign_allowed_on_groups.split("|");
```
tgxworld: Does `this.siteSettings` not work here?

jobs/scheduled/enqueue_reminders.rb
```diff
@@ -17,14 +17,19 @@ def skip_enqueue?
...
+      allowed_groups = SiteSetting.assign_allowed_on_groups.split('|')
 
       TopicCustomField
         .joins(<<~SQL
           LEFT OUTER JOIN user_custom_fields ON topic_custom_fields.value::INT = user_custom_fields.user_id
           AND user_custom_fields.name = '#{PendingAssignsReminder::REMINDED_AT}'
         SQL
         ).joins("INNER JOIN users ON topic_custom_fields.value::INT = users.id")
-        .where("users.moderator OR users.admin")
+        .where('users.id IN (
```
tgxworld: Let's do a `JOIN` here instead. `IN` clauses are generally quite inefficient because it becomes an array comparison instead of a potential index scan.

config/settings.yml
```diff
@@ -27,3 +27,11 @@ plugins:
     client: true
     default: 10
     min: 1
+  assign_allowed_on_groups:
```
tgxworld: I'm not sure about this as a site setting and I feel like this should be a group option stored in a group custom fields. The trouble with a site setting is that when a group is renamed or deleted, the site settings becomes inaccurate. Having to handle that in a nice manner isn't going to be easy.

lib/topic_assigner.rb
```diff
@@ -90,32 +90,30 @@ def self.auto_assign(post, force: false)
   end
 
   def self.is_last_staff_post?(post)
+    allowed_user_ids = User.assign_allowed.pluck(:id).join(',')
+
     sql = <<~SQL
       SELECT 1 FROM posts p
-       JOIN users u ON u.id = p.user_id AND (moderator OR admin)
+       JOIN users u ON u.id = p.user_id AND u.id IN (#{allowed_user_ids})
```
tgxworld: Let's try doing a join first if possible.

config/settings.yml
romanrizzi: I think that the site setting makes it easier to manage multiple allowed groups. Maybe we can track when a group gets renamed/deleted and make sure it remains synced?

assets/javascripts/discourse/services/task-actions.js.es6
romanrizzi: Nope, that's why I had to use this 😨

jobs/scheduled/enqueue_reminders.rb
romanrizzi: I changed it to a join but I'll fetch allowed group ids in another query, don't think it's a major performance hit.

config/settings.yml
tgxworld: Hmm maybe but I still think that having to maintain this sync is quite tricky. Maybe have a look how [type: category](https://github.com/discourse/discourse/blob/e2444e0d31588e373a018b5321b40cd3c56d7267/config/site_settings.yml#L674-L676) for site settings is handled? I'm not sure myself but worth exploring.

romanrizzi: Ended up using `type: group_list` instead.

romanrizzi:
Quick update on this one: I changed the code to track when a group gets renamed/deleted and update the setting with a model callback. I did some research and this was the only way.

I plan to merge it tomorrow morning. cc @tgxworld

PR #34, opened 2019-05-09
Disable flagging behaviour in favor of the new reviewable API

romanrizzi:
This PR bypasses the callbacks defined for `assign_locks_flags` and `flags_require_assign` since the queue was deprecated in favor of the new Reviewable API.

There's no need to change client code since the connectors will never be used (the queue was removed). However, we'll leave them there for backward compatibility reasons.


romanrizzi:
@eviltrout Reminder set 👍

PR #36, opened 2019-06-11
FEATURE: Staff is always allowed to assign

romanrizzi:


jobs/scheduled/enqueue_reminders.rb
```diff
@@ -17,9 +17,18 @@ def skip_enqueue?
 
     def allowed_group_ids
       allowed_groups = SiteSetting.assign_allowed_on_groups.split('|')
+      return [] if allowed_groups.empty?
+
       Group.where(name: allowed_groups).pluck(:id).join(',')
```
eviltrout: I would expect a method called `allowed_group_ids` to return a collection of ids, not a string. I think it would be nicer if you returned the `pluck` here and had the `membership_condition` method call `join`

jobs/scheduled/enqueue_reminders.rb
```diff
@@ -17,9 +17,18 @@ def skip_enqueue?
...
     end
 
+    def membership_condition
+      group_ids = allowed_group_ids
+      condition = 'users.admin OR users.moderator'
+      condition += " OR group_users.group_id IN (#{group_ids})" unless group_ids.empty?
```
eviltrout: `if group_ids.present?` is easier to understand than the double negative of `unless / empty`

PR #37, opened 2019-06-24
FIX: Staff can always assign

romanrizzi:
The first commit only brings the feature back. The second commit fixes a bug where the assign modal was not letting us search for users correctly.

cc @SamSaffron

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -49,6 +49,8 @@ after_initialize do
   reviewable_api_enabled = defined?(Reviewable)
 
   add_to_class(:user, :can_assign?) do
+    return true if staff?
+
     @can_assign ||=
       begin
         allowed_groups = SiteSetting.assign_allowed_on_groups.split('|').compact
@@ -62,11 +64,18 @@ after_initialize do
 
   add_class_method(:user, :assign_allowed) do
     allowed_groups = SiteSetting.assign_allowed_on_groups.split('|')
-    where('users.id IN (
-      SELECT user_id FROM group_users
-      INNER JOIN groups ON group_users.group_id = groups.id
-      WHERE groups.name IN (?)
-    )', allowed_groups)
+
+    if allowed_groups.present?
+      real.where('
+        users.admin OR users.moderator OR
+        users.id IN (
+          SELECT user_id FROM group_users
+          INNER JOIN groups ON group_users.group_id = groups.id
+          WHERE groups.name IN (?)
+        )', allowed_groups)
+    else
+      real.where('users.admin OR users.moderator')
+    end
   end
 
   add_model_callback(Group, :before_update) do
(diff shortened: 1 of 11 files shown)


plugin.rb
```diff
@@ -60,13 +62,27 @@
 
   add_to_class(:guardian, :can_assign?) { user && user.can_assign? }
 
+  add_to_serializer(:site, :assign_allowed_on_groups) do
+    staff_group = Group.select(:name).find_by(id:Group::AUTO_GROUPS[:staff])
```
SamSaffron: I am so uneasy having this query every we load the page. 

Maybe instead this site setting just had ids in it?

https://github.com/discourse/discourse/blob/5bc5c02af6877d0aed5ccdadf7bcf5f70385887e/app/models/group.rb#L70-L70

Staff has the group id 3, so you can start with that as the default. A bunch of stuff cleans up internally once that is in play.

romanrizzi:
@SamSaffron I updated the PR based on your comment.

I had considered your idea before starting working on this PR but I don't think there's an easy way of switching the site setting to store the IDs without changing core 🤔. What I propose is to send the allowed groups list alongside the assign suggestions since they're only used by the assign modal.

spec/requests/assign_controller_spec.rb
```diff
@@ -65,10 +65,31 @@
       TopicAssigner.new(post.topic, user).assign(another_admin)
 
       get '/assign/suggestions.json'
-      suggestions = JSON.parse(response.body).map { |u| u['username'] }
+      suggestions = JSON.parse(response.body)['suggested_users'].map { |u| u['username'] }
 
       expect(suggestions).to contain_exactly(user.username)
     end
+
+    it 'xxxxx' do
```
ZogStriP: Might want a better description here ;)

spec/requests/assign_controller_spec.rb
```diff
@@ -65,10 +65,31 @@
...
+      staff_group = Group.select(:name).find_by(id: Group::AUTO_GROUPS[:staff])
+
+      get '/assign/suggestions.json'
+      allowed_groups = JSON.parse(response.body).fetch('assign_allowed_groups')
+
+      expect(allowed_groups).to contain_exactly(staff_group.name)
+    end
+
+
+    it 'xsxsxsxs' do
```
ZogStriP: Ditto ^^

spec/requests/assign_controller_spec.rb
romanrizzi: ooops 🤣

SamSaffron:
> I had considered your idea before starting working on this PR but I don't think there's an easy way of switching the site setting to store the IDs without changing core

I say start off by changing core then, even just giving it one more option. I far prefer a cleaner implementation even if it means that our building blocks need to be re-designed.

romanrizzi:
I did some research and changing the site setting to store IDs instead of group names will be the best approach. This setting type is also used by the ads plugin but can be safely changed.

I'll make the necessary changes and will open a new PR.

PR #38, opened 2019-07-04
DEV: use group ids to allow assign on groups

romanrizzi:
The build is broken because this depends on https://github.com/discourse/discourse/pull/7860

This PR let us use group IDs as default values for the `assign_allowed_on_groups` setting.

To maintain backwards compatibility, we also ship:

- A way to migrate the setting values only after being sure the core's code is available.
- A way to keep the feature working when we still have group names instead of IDs

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -51,7 +51,15 @@ after_initialize do
   # TODO: Remove this once 2.4 becomes the new stable.
   current_version = ActiveRecord::Migrator.current_version
   min_version = 201_907_081_533_31
-  attribute = current_version >= min_version ? 'id' : 'name'
+  above_min_version = current_version >= min_version
+
+  # Dinamically sets the default value, supports older versions.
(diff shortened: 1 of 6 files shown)


plugin.rb
```diff
@@ -48,11 +48,24 @@
 =end
   reviewable_api_enabled = defined?(Reviewable)
 
+
+  # We need to migrate existing settings automatically
+  # TODO: remove when 2.4 becomes the new stable
+  group_list_changed = (SiteSetting.group_list_changed rescue false)
```
SamSaffron: You are following the same pattern in multiple places, just define a constant in plugin.rb instead. then you don't need to recheck this or handle another exception. 

Also... instead of this, SiteSetting.respond_to? :group_list_changed ?

SamSaffron:
Overall I love this change, do some minor cleanup by moving the "feature detection" into a constant and using respond_to ... then go ahead and merge

eviltrout:
We should hold off until we sort out the strategy for this in master before merging.

PR #39, opened 2019-07-07
FIX: Skip enqueuing reminders if no groups are allowed.

udan11:
With an empty SiteSetting.assign_allowed_on_groups, it used to generate
an invalid query containing "group_users.group_id IN ()".

### DIFF
--- a/jobs/scheduled/enqueue_reminders.rb
+++ b/jobs/scheduled/enqueue_reminders.rb
@@ -12,7 +12,7 @@ module Jobs
     private
 
     def skip_enqueue?
-      SiteSetting.remind_assigns_frequency.nil? || !SiteSetting.assign_enabled?
+      SiteSetting.remind_assigns_frequency.nil? || !SiteSetting.assign_enabled? || SiteSetting.assign_allowed_on_groups.blank?
     end
 
     def allowed_group_ids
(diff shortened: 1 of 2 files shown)


jobs/scheduled/enqueue_reminders.rb
```diff
@@ -12,7 +12,7 @@ def execute(_args)
     private
 
     def skip_enqueue?
-      SiteSetting.remind_assigns_frequency.nil? || !SiteSetting.assign_enabled?
+      SiteSetting.remind_assigns_frequency.nil? || !SiteSetting.assign_enabled? || !SiteSetting.assign_allowed_on_groups.present?
```
eviltrout:
```suggestion
      SiteSetting.remind_assigns_frequency.nil? || !SiteSetting.assign_enabled? || SiteSetting.assign_allowed_on_groups.blank?
```

I find `blank?` reads easier than `!present?`

ZogStriP:
Can we also get a test?

spec/jobs/scheduled/enqueue_reminders_spec.rb
```diff
@@ -16,6 +16,14 @@
       assert_reminders_enqueued(0)
     end
 
+    it 'does not enqueue remidners when no groups are allowed to assign' do
```
ZogStriP:
```suggestion
    it 'does not enqueue reminders when no groups are allowed to assign' do
```

PR #41, opened 2019-07-17
 DEV: use group ids to allow assign on groups [Undo Revert] 

romanrizzi:
This PR reverts the revert of #38.

### DIFF
--- a/db/migrate/20190718144722_set_assign_allowed_on_groups_default.rb
+++ b/db/migrate/20190718144722_set_assign_allowed_on_groups_default.rb
@@ -4,7 +4,7 @@ class SetAssignAllowedOnGroupsDefault < ActiveRecord::Migration[5.2]
   def up
     current_values = DB.query_single("SELECT value FROM site_settings WHERE name = 'assign_allowed_on_groups'").first
 
-    # Dinamically sets the default value, supports older versions.
+    # Dynamically sets the default value, supports older versions.
     if current_values.nil?
       min_version = 201_907_171_337_43
       migrated_site_setting = DB.query_single(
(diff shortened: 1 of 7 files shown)


db/migrate/20190717132703_set_assign_allowed_on_groups_default.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class SetAssignAllowedOnGroupsDefault < ActiveRecord::Migration[5.2]
+  def up
+    current_values = SiteSetting.assign_allowed_on_groups.split('|')
+    
+    # Dinamically sets the default value, supports older versions.
+    if current_values.empty?
```
eviltrout: We discussed this interally on chat but this needs to be modified to use SQL.

romanrizzi:
@eviltrout  I did some more changes to this so we can the plugin works alongside tests-passed/master/beta in all the cases. Can you take a look at my last commit?

db/migrate/20190718144722_set_assign_allowed_on_groups_default.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class SetAssignAllowedOnGroupsDefault < ActiveRecord::Migration[5.2]
+  def up
+    current_values = DB.query_single("SELECT value FROM site_settings WHERE name = 'assign_allowed_on_groups'").first
+
+    # Dinamically sets the default value, supports older versions.
```
eviltrout:
```suggestion
    # Dynamically sets the default value, supports older versions.
```

PR #45, opened 2019-07-22
FIX: System user should be allowed to use plugin.

nbianca:
This makes it easier to retrieve assigned list through API.


plugin.rb
```diff
@@ -51,6 +51,7 @@
   add_to_class(:user, :can_assign?) do
     @can_assign ||=
       begin
+        return true if self == Discourse.system_user
```
eviltrout: Is there a reason this is preferable over an `admin?` check?

PR #46, opened 2019-09-02
FEATURE: New rules for assigning a topic.

romanrizzi:
As discussed [here](https://meta.discourse.org/t/bug-possible-for-admin-to-assign-pm-to-moderator-who-does-not-have-access-to-the-pm/116571/6), it shouldn't be possible to assign a topic to a user when:

- The user won't be able to see the topic.
- The user is not a member of the assign allowed users.

This PR adds some backend rules to prevent these scenarios.


SamSaffron:
This looks good to me, merge first thing when you start your day!

PR #47, opened 2019-09-02
DEV: Remove dreprecated review queue code

romanrizzi:


CvX:
@romanrizzi `assign_locks_flags` and `flags_require_assign` were removed from config/settings.yml in this PR, but both are still used in the code. Backend tests on https://github.com/discourse/discourse are failing because of this line:

https://github.com/discourse/discourse-assign/blob/master/spec/integration/assign_spec.rb#L77

romanrizzi:
I wonder why CI didn’t fail earlier. I merged this PR several hours ago.

I’m currently AFK so I can’t fix it right now

romanrizzi:
Thanks for letting me know, Jarek. I’ll take a look as soon as I can.

CvX:
> I wonder why CI didn’t fail earlier.

The failing tests has is:

```ruby
it "do not raise error if topic is deleted", unless: defined?(Reviewable) do
```

Here, `Reviewable` is clearly defined (I can see in Travis logs that 4 out of 5 specs were run in this file). But in the context of https://github.com/discourse/discourse `Reviewable` doesn't exist?

romanrizzi:
Reviewable is always defined since it’s present on both stable and
tests-passed (Wasn’t defined on Discourse 2.2).

This is a travis only failure? We’re running the suite against older
Discourse versions?

CvX:
Ok, turns out that `Reviewable` in the Discourse context is undefined when tests are being loaded, but gets defined later, when these tests are run.

Would be good to know exactly why the difference in code loading order, but in the meantime a simplest fix/workaround would be:

```diff
    before do
+     skip if defined?(Reviewable)
      SiteSetting.assign_locks_flags = true

      TopicCustomField.create!(
        topic_id: post.topic_id,
        name: TopicAssigner::ASSIGNED_TO_ID,
        value: Discourse.system_user.id
      )
    end

    it "do not raise error if topic is deleted", unless: defined?(Reviewable) do
```

PR #50, opened 2019-09-13
FIX: do not escape `fancy_title` again.

xrav3nz:
> CONTEXT: https://meta.discourse.org/t/quick-access-to-bookmarks-and-messages-on-user-menu/32787/50

This relies on the `escapedContent` attribute introduced in https://github.com/discourse/discourse/pull/8095.

Screen Shot 2019-09-12 at 9 33 49 PM

Thanks!


eviltrout:
Thanks!

PR #49, opened 2019-09-08
FEATURE: Quick access panel for assignments.

xrav3nz:
> CONTEXT: [Quick access to bookmarks and messages on user menu](https://meta.discourse.org/t/quick-access-to-bookmarks-and-messages-on-user-menu/32787/12?u=xrav3nz)
> Related Core PR: https://github.com/discourse/discourse/pull/8073

Open to suggestions, thanks!

PR #51, opened 2019-09-13
TEST: Acceptance test for quick access assignments

xrav3nz:
This is a test-only change for the "quick access assignments" feature introduced in https://github.com/discourse/discourse-assign/pull/49.

Thanks!


eviltrout:
Thanks, looks good.

PR #53, opened 2019-10-26
DEV: Use BasicUserSerializer for assigned user

nbianca:


SamSaffron:
Are you sure we are getting all the custom fields we need? Can you expand on this a bit?

nbianca:
I was working on a project where initially I needed the custom fields to fetch extra info for the user. It turned out that I did not really need it and since nobody really uses this (because it was recently added and the pattern is quite odd), I decided to remove it.

SamSaffron:
Cool, thanks!

PR #54, opened 2019-10-27
DEV: Avoid useless join

nbianca:


app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -98,8 +98,6 @@ def assigned
 
       users = User
         .where("users.id IN (SELECT value::int FROM topic_custom_fields WHERE name = 'assigned_to_id' AND topic_id IN (?))", topics.map(&:id))
-        .joins('join user_emails on user_emails.user_id = users.id AND user_emails.primary')
-        .select(AvatarLookup.lookup_columns)
```
SamSaffron: This is selecting reduced columns here though, not sure it is 100% useless. Also it is looking for real users. 

Maybe `User.real` is a better thing to do here?

PR #57, opened 2019-11-19
FEATURE: Improve assign mailer site setting

udan11:


app/models/assign_mailer_site_settings.rb
```diff
@@ -0,0 +1,22 @@
...
+class AssignMailerSiteSettings < EnumSiteSetting
+
+  def self.valid_value?(val)
+    values.any? { |v| v[:value].to_s == val.to_s }
+  end
+
+  def self.values
+    @values ||= [
+      { name: 'discourse_assign.assign_mailer.never', value: 'never' },
+      { name: 'discourse_assign.assign_mailer.different', value: 'different' },
```
SamSaffron: what does "different" mean?

udan11: It means that the assign mailer will be send an email only if the assigner and assignee are different. Regis recommended me to change this to `different_users`, so I did it. I hope everything is clear now.

lib/topic_assigner.rb
```diff
@@ -189,7 +189,7 @@ def assign(assign_to, silent: false)
       )
     end
 
-    if SiteSetting.assign_mailer_enabled
+    if SiteSetting.assign_mailer == 'always' || (SiteSetting.assign_mailer == 'different_users' && @assigned_by.id != assign_to.id)
```
ZogStriP: I've come to hate magic strings. Can we instead have an enum that is used everywhere? (like we do for `Post.types` for example)

PR #59, opened 2019-12-19
FIX: Update topic-list-item to reflect changes in core

oblakeerickson:
This PR in discourse core:

https://github.com/discourse/discourse/pull/8589

makes a breaking change in this plugin. This is a fix for it.

### DIFF
--- a/assets/javascripts/discourse/components/assigned-topic-list-item.js.es6
+++ b/assets/javascripts/discourse/components/assigned-topic-list-item.js.es6
@@ -1,5 +1,5 @@
-import { ListItemDefaults } from "discourse/components/topic-list-item";
+import TopicListItem from "discourse/components/topic-list-item";
 
-export default Ember.Component.extend(ListItemDefaults, {
+export default TopicListItem.extend({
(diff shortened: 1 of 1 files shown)


oblakeerickson:
These tests are likely failing, because the changes in core haven't been applied yet.

eviltrout:
@oblakeerickson if that's the case, does that mean if we merge this it'll break people on the beta/stable branches?

oblakeerickson:
@eviltrout correct. We will need to wait to merge this until the PR in core is merged in.

eviltrout:
@oblakeerickson is there any way we could make it backwards compatible?

oblakeerickson:
I'll give it a shot and let you know.

oblakeerickson:
I was able to add a backwards compatible fix which seems to be working fine locally. Something weird is going on with the qunit tests and they seem to be hanging in a weird spot. I'll keep investigating and see if I can figure out what is actually going on.

eviltrout:
Thanks for the update @oblakeerickson ! sounds good

ZogStriP:
@oblakeerickson should this be merged?

PR #61, opened 2020-02-28
FEATURE: Trigger webhook when assigning and unassigning topics

oblakeerickson:
This PR creates a custom webhook event that you can now select when
creating a webhook to trigger only when a topic has been assigned or
unassigned.

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -35,12 +35,6 @@ after_initialize do
   require 'topic_assigner'
   require 'pending_assigns_reminder'
 
-  class ::WebHookEventType
-    ASSIGN = 101.freeze
-  end
-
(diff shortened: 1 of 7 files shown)


plugin.rb
```diff
@@ -35,6 +35,12 @@
   require 'topic_assigner'
   require 'pending_assigns_reminder'
 
+  class ::WebHookEventType
+    ASSIGN = 101.freeze
```
eviltrout: This is a bit difficult to swallow because it means plugins could use the same id.

I know in the past @SamSaffron suggested plugins make PRs to core to reserve ids. It's not great but might be better than this? Sam do you still think that's the best approach?

PR #65, opened 2020-04-23
REFACTOR: various refactorings and cleanings

jjaffeux:


assets/javascripts/discourse-assign/controllers/assign-user.js.es6
```diff
@@ -22,38 +27,38 @@ export default Ember.Controller.extend({
...
-    assign() {
-      let path = "/assign/assign";
-
-      if (Ember.isEmpty(this.get("model.username"))) {
-        path = "/assign/unassign";
-        this.set("model.assigned_to_user", null);
-      }
+  @action
+  assignUser(user) {
+    this.setProperties({
```
eviltrout, line 32: Minor but why not this.model.setProperties?

davidtaylorhq:
@jjaffeux it looks like this has become quite outdated. Do you think we should try to resolve the conflicts? Or will it be easier to close and start again?

jjaffeux:
totally forgot about this, feel free to start from scratch, maybe look at it so try to search for similar things

PR #68, opened 2020-06-17
Changed forbidden_assign_to string

Ahmedgagan:


eviltrout:
I don't agree with this change. There are several reasons why a user might not be able to see a topic and this changes the error to be too specific.

PR #70, opened 2020-06-26
Group Assignments Summary

Ahmedgagan:
New feature: group-assignments-summary


app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -114,8 +115,75 @@ def assigned
       render json: { topics: serialize_data(topics, AssignedTopicSerializer) }
     end
 
+    def group_assignments
+      if SiteSetting.assign_enabled?
```
davidtaylorhq, line 119: Rather than wrapping this all in a huge `if` statement, you could do

```
raise Discourse::InvalidAccess.new if !SiteSetting.assign_enabled?
```
(Also, maybe we should use Discourse::NotFound, since 404 is a better error code for this situation)

app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -114,8 +115,75 @@ def assigned
...
+            .where("tcf.value::int = ?", users[0].id)
+            .order('tcf.value::integer, topics.bumped_at desc')
+
+          load_more = topics.to_a.length > 30 * (params[:page].to_i + 1)
+
+          topics = topics
+            .offset(offset)
+            .limit(30)
+
+        end
```
davidtaylorhq, line 154: There is lots of repetition in the `if` and `else` part of this statement. Can we move some stuff outside the if statement?

app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -114,8 +115,75 @@ def assigned
...
+
+          load_more = topics.to_a.length > 30 * (params[:page].to_i + 1)
+
+          topics = topics
+            .offset(offset)
+            .limit(30)
+
+          users = User.where("users.id IN (SELECT group_users.user_id FROM group_users WHERE (group_id IN (SELECT id FROM groups WHERE name = ?)))", params[:group_name])
+        else
+          users = User.where("username = ?", params[:group_name])
```
davidtaylorhq, line 139: Why are we using the `group_name` parameter to describe a user? That is kind of confusing.

When requesting the topics for a single user, do we need to use this endpoint? Or could we re-use the existing per-user endpoint?

assets/javascripts/discourse-assign/controllers/group-assignments.js.es6
```diff
@@ -0,0 +1,6 @@
+import { inject as service } from "@ember/service";
+import Controller from "@ember/controller";
+export default Controller.extend({
+  router: service(),
+  queryParams: ["route_type"]
```
davidtaylorhq, line 5: What does this parameter do? Can we name it something more descriptive?

assets/javascripts/discourse-assign/routes/group-assignments.js.es6
```diff
@@ -0,0 +1,18 @@
+import { ajax } from "discourse/lib/ajax";
+import Route from "@ember/routing/route";
+
+export default Route.extend({
+  model() {
+    return ajax(`/groups/${this.modelFor("group").get("name")}/members.json`, {
```
davidtaylorhq, line 6: What happens if a group has lots (e.g. 50+) members. Does this still work?

assets/javascripts/discourse/components/group-reports-nav-item.js.es6
```diff
@@ -0,0 +1,34 @@
...
+          }
+        });
+      }
+      this.set(
+        "hasAssignments",
+        render &&
+          this.siteSettings.assign_enabled &&
+          response.topic_list.topics.length > 0
+      );
+    });
```
davidtaylorhq, line 27: We don't really want to be making ajax requests while rendering components. Let's simplify this to match [the per-user behaviour](https://github.com/discourse/discourse-assign/blob/f312ece4a97a7bf40c1d3a734b754df1291758dc/assets/javascripts/discourse-assign/connectors/user-activity-bottom/assigned-list.hbs#L1).

If `current_user.can_assign` is true, show the tab. If not, then hide it.

app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -2,8 +2,9 @@
 
 module DiscourseAssign
   class AssignController < ApplicationController
+    include TopicListResponder
     requires_login
-    before_action :ensure_logged_in, :ensure_assign_allowed
+    before_action :ensure_logged_in, :ensure_assign_allowed, :group_assignments
```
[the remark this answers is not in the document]
Ahmedgagan: ooh, this was added by mistake. will remove it in next commit😅

app/controllers/discourse_assign/assign_controller.rb
Ahmedgagan: We need to use this endpoint as existing one doesn't allow getting data of other users unless current_user is admin

app/controllers/discourse_assign/assign_controller.rb
Ahmedgagan: Changed as specified😀

app/controllers/discourse_assign/assign_controller.rb
Ahmedgagan: Sure, it's now done as specified in the new commit😀

assets/javascripts/discourse-assign/controllers/group-assignments.js.es6
Ahmedgagan: I just removed it, as it wasn't useful

assets/javascripts/discourse-assign/routes/group-assignments.js.es6
Ahmedgagan: Yes, in the new commit it works with 50+ users too.

assets/javascripts/discourse/components/group-reports-nav-item.js.es6
Ahmedgagan: Sure, it's now done as specified in the new commit😀

app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -114,8 +115,74 @@ def assigned
       render json: { topics: serialize_data(topics, AssignedTopicSerializer) }
     end
 
+    def assignments_count
```
davidtaylorhq, line 118: Let's try to avoid another AJAX request to load the assignments count. We should be able to include it in the group serializer

Something like `add_to_serializer(:group, :assignments_count)` in plugin.rb

app/controllers/discourse_assign/assign_controller.rb
davidtaylorhq: Oh I think that is probably a mistake. The assign plugin was originally for staff-only, but now we made it so you can customize the groups. So instead of checking admin, can you change that route to check for `current_user.can_assign`?

plugin.rb
```diff
@@ -36,6 +38,12 @@
   add_to_serializer(:user, :reminders_frequency) do
     RemindAssignsFrequencySiteSettings.values
   end
+
+  add_to_serializer(:group_show, :assignment_count) do
+    Topic.joins("JOIN topic_custom_fields tcf ON topics.id = tcf.topic_id AND tcf.name = 'assigned_to_id' AND tcf.value IS NOT NULL")
+      .where("tcf.value IN (SELECT group_users.user_id::varchar(255) FROM group_users WHERE (group_id IN (SELECT id FROM groups WHERE name = ?)))", object.name).count
+  end
+
```
davidtaylorhq, line 46: Looks good :+1:

We just need to make sure it is secure. This should only be included for users where `scope.can_assign?` is true. (see [here](https://github.com/discourse/discourse-assign/blob/master/plugin.rb#L237-L244) for a similar implementation)

app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -114,8 +115,64 @@ def assigned
       render json: { topics: serialize_data(topics, AssignedTopicSerializer) }
     end
 
+    def group_assignments
```
davidtaylorhq, line 118: A lot of this code seems similar to `def assigned` above. Can we share any logic between them?

app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -2,6 +2,7 @@
 
 module DiscourseAssign
   class AssignController < ApplicationController
+    include TopicListResponder
```
davidtaylorhq, line 5: I think we can remove this now?

app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -116,6 +117,25 @@ def assigned
 
     private
 
+    def display_topic_list(topics, users, more_topics_url)
```
davidtaylorhq, line 120: Can we remove this function?

plugin.rb
davidtaylorhq: ^^ I think this comment still needs to be addressed?

plugin.rb
```diff
@@ -186,19 +196,58 @@ class ::TopicListItemSerializer
...
+    list = default_results.where("
       topics.id IN (
         SELECT topic_id FROM topic_custom_fields
         WHERE name = 'assigned_to_id'
         AND value = ?)
     ", user.id.to_s)
       .order("topics.bumped_at DESC")
 
-    create_list(:assigned, {}, list)
+    return create_list(:assigned, {}, list)
```
davidtaylorhq, line 210: No need to have `return` here in Ruby

(I think rubocop will actually fail for this)

plugin.rb
```diff
@@ -186,19 +196,58 @@ class ::TopicListItemSerializer
...
+  end
+
+  add_to_class(:list_controller, :messages_assigned) do
+    page = (params[:page].to_i || 0).to_i
+
+    user = User.find_by("LOWER(username) = ?", params[:username])
```
davidtaylorhq, line 216: We have a special column here which is indexed. Use `username_lower` rather than `LOWER(username)`

Or even better, use `User.find_by_username(params[:username])` and it will happen automatically :D

plugin.rb
```diff
@@ -186,19 +196,58 @@ class ::TopicListItemSerializer
...
+  add_to_class(:topic_query, :list_group_messages_assigned) do |group|
+    list = default_results.where("
+      topics.id IN (
+        SELECT topic_id FROM topic_custom_fields
+        WHERE name = 'assigned_to_id'
+        AND value IN (SELECT user_id::varchar(255) from group_users where group_id = ?))
+    ", group.id.to_s)
+      .order("topics.bumped_at DESC")
+
+    return create_list(:assigned, {}, list)
```
davidtaylorhq, line 236:
```suggestion
    create_list(:assigned, {}, list)
```

plugin.rb
```diff
@@ -36,6 +39,21 @@
...
   end
+
+  add_to_serializer(:group_user, :assignments_count) do
+    if scope.can_assign?
+      Topic.joins("JOIN topic_custom_fields tcf ON topics.id = tcf.topic_id AND tcf.name = 'assigned_to_id' AND tcf.value IS NOT NULL")
+        .where("tcf.value = ?", object.id.to_s).count
+    end
+  end
+
+  add_to_serializer(:group_show, :assignment_count) do
```
davidtaylorhq, line 50: We still need to secure this. It should only be visible to users who `can_assign`

plugin.rb
```diff
@@ -186,11 +199,10 @@ class ::TopicListItemSerializer
   require_dependency 'list_controller'
   class ::ListController
     generate_message_route(:private_messages_assigned)
-    generate_message_route(:messages_assigned)
   end
 
   add_to_class(:topic_query, :list_messages_assigned) do |user|
-    list = joined_topic_user.where("
+    list = default_results.where("
```
davidtaylorhq, line 205: (I know I asked you to use `default_results` originally, sorry 😅 )

I realised that `default_results` doesn't include PMs. We like to have a combined list of 'topics and PMs' for the assigned list. So using `joined_topic_user` worked well for that in the past. 

However... now that we are allowing non-admins to see this list, we need to add some security. I think something like this could work (and also the same on the `group_messages_assigned` method) 

Please test this though - I haven't checked it

```suggestion
    secure = Topic.listable_topics.secured(@guardian).or(Topic.private_messages_for_user(@user))
    list = joined_topic_user.where(secure).where("
```

plugin.rb
```diff
@@ -20,7 +20,10 @@
 Discourse::Application.routes.append do
   mount ::DiscourseAssign::Engine, at: "/assign"
   get "topics/private-messages-assigned/:username" => "list#private_messages_assigned", as: "topics_private_messages_assigned", constraints: { username: ::RouteFormat.username }
-  get "topics/messages-assigned/:username" => "list#messages_assigned", as: "topics_messages_assigned", constraints: { username: ::RouteFormat.username }
+  get "/topics/messages-assigned/:username" => "list#messages_assigned", constraints: { username: ::RouteFormat.username }
+  get "/topics/group-messages-assigned/:groupname" => "list#group_messages_assigned", constraints: { username: ::RouteFormat.username }
```
davidtaylorhq, line 24: group-messages-assigned is kind of confusing, because 'messages' implies 'private messages'.

Please can we rename it to `group-topics-assigned`? And also rename the matching method names.

The old `messages-assigned` route can stay for now, we'll look at changing it in the future.

davidtaylorhq:
@Ahmedgagan looks like we still have a couple of qunit failures in Travis

PR #71, opened 2020-06-28
FIX: Don't set assign button's height

CvX:
"we can't use flex for buttons yet" – now we can, and we do! https://github.com/discourse/discourse/commit/e71f5e8951bcffdaac3476d96efaa2232b809a4c

Fixes an issue with assign button height in some themes.

Before/after:

image
image

Core theme, no change:

image
Screen Shot 2020-06-28 at 18 25 19

### DIFF
--- a/assets/stylesheets/assigns.scss
+++ b/assets/stylesheets/assigns.scss
@@ -81,10 +81,7 @@
   }
 }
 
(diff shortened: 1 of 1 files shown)


coding-horror:
💪🎉

PR #75, opened 2020-07-13
display name changed as per site settings

Ahmedgagan:


assets/javascripts/discourse/components/group-assignments-filter.js.es6
```diff
@@ -1,5 +1,15 @@
...
+  displayName: "",
+
+  init(){
+    if(this.siteSettings.prioritize_username_in_ux){
+      this.set("displayName",this.filter.username);
+    }else{
+      this.set("displayName",this.filter.displayName);
+    }
+    this._super(...arguments);
+  }
```
davidtaylorhq, line 14: I think a better approach here would be to use a computed property. If you search core for `@discourseComputed` you'll find some examples. [For example](https://github.com/discourse/discourse/blob/fbf29443bdb84e0a07a34a82d7893ab967854bd9/app/assets/javascripts/admin/components/admin-report-table-cell.js#L11-L14).

Ahmedgagan: Solved this in new commit

davidtaylorhq: This should be computed based on all the properties it uses. So `filter.username` and `filter.displayName` should also be included in the `discourseComputed(...)`

Ahmedgagan: Sure, just added these too in discourseComputed() in latest commit

davidtaylorhq: And then we can pass them as variables into the function, just like you did for the site setting :)

Ahmedgagan: Oooh sorry for that😅😅

Ahmedgagan: Updated this one😄

assets/javascripts/discourse/components/group-assignments-filter.js.es6
```diff
@@ -1,5 +1,14 @@
+import discourseComputed from "discourse-common/utils/decorators";
 import Component from "@ember/component";
 
 export default Component.extend({
-  tagName: "li"
+  tagName: "li",
+
+  @discourseComputed("siteSettings.prioritize_username_in_ux", "filter.username", "filter.displayName")
```
davidtaylorhq, line 7: Where does `filter.displayName` come from? Should that be `filter.name`?

Apart from that question, this is looking great :D

Ahmedgagan: filter.name property doesn't value when the user doesn't set name property in the signup, but displayName always have the value

PR #80, opened 2020-07-17
Members

Ahmedgagan:
Custom route for group_members


app/serializers/group_user_assigned_serializer.rb
```diff
@@ -0,0 +1,22 @@
...
+class GroupUserAssignedSerializer < BasicUserSerializer
+  include UserPrimaryGroupMixin
+
+  attributes :name,
+             :title,
+             :last_posted_at,
+             :last_seen_at,
+             :added_at,
+             :assignments_count,
+             :timezone
```
davidtaylorhq, line 12: Do we need all these attributes? `name` is already included in BasicUserSerializer. And I don't think we're actually using title/last_posted_at/last_seen_at/added_at/timezone for the assign UI?

assets/javascripts/discourse/templates/components/group-assignments-filter.hbs
```diff
@@ -1,6 +1,6 @@
 {{#if show-avatar}}
   {{#link-to "group.assignments.show" filter.username_lower}}
-    {{avatar filter avatarTemplatePath="avatar_template" usernamePath="username" imageSize="small"}} {{displayName}}
+    {{avatar filter avatarTemplatePath="avatar_template" usernamePath="username" imageSize="small"}} {{displayName}} {{concat '(' filter.assignments_count concat ')' }}
```
davidtaylorhq, line 3: How about `({{filter.assignments_count}})`? I don't think we need concat here

app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -122,6 +122,28 @@ def assigned
...
+      group = Group.find_by(name: params[:group_name])
+      members = User
+        .joins("LEFT OUTER JOIN group_users g on users.id=g.user_id LEFT OUTER JOIN user_options uo on uo.user_id=users.id LEFT OUTER JOIN topic_custom_fields tcf ON tcf.value::int = users.id")
+        .where("tcf.name = 'assigned_to_id' AND g.group_id=? AND (users.id > 0)", group.id)
+        .order('COUNT(users.id) DESC')
+        .group('users.id, added_at, uo.timezone')
+        .select('users.*, COUNT(users.id) as "assignments_count", g.created_at as "added_at", uo.timezone')
+        .limit(limit)
+        .offset(offset)
+        .includes(:primary_group)
```
davidtaylorhq, line 142: See comments on the serializer. I think we might be able to remove a lot of this (e.g. primary group, timezone, added_at`)

app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -122,6 +122,28 @@ def assigned
...
+      limit = (params[:limit] || 50).to_i
+      offset = params[:offset].to_i
+
+      raise Discourse::InvalidParameters.new(:limit) if limit < 0 || limit > 1000
+      raise Discourse::InvalidParameters.new(:offset) if offset < 0
+      raise Discourse::NotFound.new if !params[:group_name].present?
+
+      group = Group.find_by(name: params[:group_name])
+      members = User
+        .joins("LEFT OUTER JOIN group_users g on users.id=g.user_id LEFT OUTER JOIN user_options uo on uo.user_id=users.id LEFT OUTER JOIN topic_custom_fields tcf ON tcf.value::int = users.id")
```
davidtaylorhq, line 135: Very nice! This should be nice and efficient for assignment counts with just one query ❤️

app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -122,6 +122,28 @@ def assigned
       render json: { topics: serialize_data(topics, AssignedTopicSerializer) }
     end
 
+    def group_members
```
davidtaylorhq, line 125: I think we need a couple of security checks here. 

`can_assign` is already handled by the `before_action` in the controller ✅ 

I think we need `guardian.ensure_can_see_group_members!(group)` as well, in case the group is private. Can you also add that check to the list of assigned topics?

app/serializers/group_user_assigned_serializer.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class GroupUserAssignedSerializer < BasicUserSerializer
+  include UserPrimaryGroupMixin
```
davidtaylorhq, line 4: Do we still need the primary group mixin?

Ahmedgagan: removed in the latest commit😅

assets/javascripts/discourse-assign/controllers/group-assignments.js.es6
```diff
@@ -1,31 +1,28 @@
...
-      return;
+    if (this.model.members.length >= this.offset + 50) {
+      this.set("loading", true);
+      this.set("offset", this.offset + 50);
+      ajax(`/assign/members/${this.groupName}?offset=${this.offset}`).then(
+        result => {
+          this.set("model.members", this.model.members.concat(result.members));
+          this.set("loading", false);
+        }
+      );
```
davidtaylorhq, line 24: Sorry I didn't mention this earlier 😓 

I don't think we should modify the `model.members` parameter. The items we are adding are not complete 'user' models, they are just simple JS objects. Can we make a `members` property in the controller, and use that instead? Make sure to reset it to an empty array each time in the `setupController` method.

Ahmedgagan: done in new commit

assets/javascripts/discourse-assign/controllers/group-assignments.js.es6
```diff
@@ -1,31 +1,33 @@
...
 export default Controller.extend({
   router: service(),
   application: controller(),
   loading: false,
+  offset: 0,
 
   findMembers(refresh) {
-    if (this.loading || !this.model) {
+    if (refresh) {
+      this.set("members", this.model.members);
```
davidtaylorhq, line 13: Do we need this? I don't think that `model.members` will have the assignment count? So maybe we should set to an empty array?

assets/javascripts/discourse-assign/controllers/group-assignments.js.es6
```diff
@@ -1,31 +1,33 @@
...
-            this.model.members.length >= this.model.user_count,
-          loading: false
-        });
-      });
+    if (this.model.members.length >= this.offset + 50) {
+      this.set("loading", true);
+      this.set("offset", this.offset + 50);
+      ajax(`/assign/members/${this.groupName}?offset=${this.offset}`).then(
+        result => {
+          this.set("members", this.members.concat(result.members));
```
davidtaylorhq, line 26: I think `members` is an ember [MutableArray](https://api.emberjs.com/ember/2.15/classes/Ember.MutableArray/methods/pushObject?anchor=pushObjects)? So we should be able to use `pushObjects`
```suggestion
          this.members.pushObjects(result.members);
```
I think that should make it more efficient for ember to re-render the template (since the array isn't being completely replaced)

assets/javascripts/discourse-assign/controllers/group-assignments.js.es6
Ahmedgagan: `model.members` have assignment count, we need to set it here as when the `setupController` will be called, `members` will be empty. To set it to members from the model, we need to do this

assets/javascripts/discourse-assign/controllers/group-assignments.js.es6
Ahmedgagan: Sure, will change this in a moment

assets/javascripts/discourse-assign/controllers/group-assignments.js.es6
davidtaylorhq: Aha! I see now. I thought `model` was a `group` model, but it's not.

Makes sense :+1:

assets/javascripts/discourse-assign/controllers/group-assignments.js.es6
Ahmedgagan: added in the latest commit

davidtaylorhq:
@Ahmedgagan looks good to me. Please can you resolve the small merge conflict?

Ahmedgagan:
> @Ahmedgagan looks good to me. Please can you resolve the small merge conflict?

Solved in latest commit

PR #81, opened 2020-07-20
More spec added

Ahmedgagan:


spec/requests/list_controller_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+    before do
+      add_to_assign_allowed_group(user)
+
+      freeze_time 1.hour.from_now
+      TopicAssigner.new(post1.topic, user).assign(user)
+
+      freeze_time 1.hour.from_now
+      TopicAssigner.new(post1.topic, user).assign(user2)
+
+      sign_in(user)
```
davidtaylorhq, line 63: Why are the freeze_time things needed here?

spec/requests/list_controller_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+    it 'returns empty user-assigned-topics-list for users not in the assigned_allowed_group' do
+      ids = []
+      get "/topics/group-topics-assigned/#{get_assigned_allowed_group_name}.json"
+      JSON.parse(response.body)['topic_list']['topics'].each do |t|
+        if t['assigned_to_user']['id'] == user2.id
+          ids.push(t['assigned_to_user']['id'])
+        end
+      end
+      expect(ids).to be_empty
+    end
```
davidtaylorhq, line 80: What is the difference between these two tests? 

The first one is checking "the list contains 1 topic, assigned to user1"

The second one is checking "the list contains no topics assigned to user2".

I think the first test covers everything? If user2 was assigned a topic, the first test would fail, so we don't need the second test?

Maybe I am mis-reading this though?

spec/requests/list_controller_spec.rb
Ahmedgagan: I've seen this (here)[https://github.com/discourse/discourse-assign/blob/68bebd77f639e16f1f002ff91d05f46f91ce61d5/spec/requests/assign_controller_spec.rb#L156], so used as it is😅😅

spec/requests/list_controller_spec.rb
Ahmedgagan: it is checking, that there are `0` assignments of the user not in this group

spec/requests/list_controller_spec.rb
davidtaylorhq: Ah I see. In that case, it was testing that the topics were ordered by 'assigned date', so having precise times was important to the test. 

I think we can safely remove it here :)

Ahmedgagan: sure, ill remove this right now.

PR #83, opened 2020-07-21
Hide tab

Ahmedgagan:
Hiding assignments-tab and assignment count to Everyone added in this PR


assets/javascripts/discourse/components/group-assignments-menu-item.js.es6
```diff
@@ -3,6 +3,11 @@ export default Ember.Component.extend({
 
   init() {
     this._super(...arguments);
-    this.set("canAssign", this.currentUser && this.currentUser.can_assign);
+    this.set(
+      "canAssign",
```
davidtaylorhq, line 7: We should rename this to "displayAssignTab" or something. Because it doesn't mean 'canAssign' any more

assets/javascripts/discourse/templates/components/group-assignments-menu-item.hbs
```diff
@@ -1,8 +1,8 @@
 {{#if canAssign}}
   <ul class ='nav-pills'>
     <li>
-      {{#link-to 'group.assignments' group}}
-        {{d-icon "user-plus" class="glyph"}}{{i18n 'discourse_assign.group_assignments'}}{{concat ' (' group.assignment_count concat ')'}}
+      {{#link-to "group.assignments" group.assignment_count}}
```
davidtaylorhq, line 4: That seems a bit weird. Did you mean to change this? I don't see why we would want to pass the count into the link-to component?

assets/javascripts/discourse/components/group-assignments-menu-item.js.es6
Ahmedgagan: Sure, ill rename this

PR #89, opened 2020-07-30
Sort - ordering on topic-list-header click

Ahmedgagan:


davidtaylorhq:
Looks great :+1: please can you add some tests to check that the ordering works for both the group and the user endpoints?

PR #91, opened 2020-08-04
Scroll top

Ahmedgagan:
Scroll-top on username-filter change


davidtaylorhq:
Another conflict here ^^

Also, just checking, does this code run when you click "back" in the browser? We don't want to reset the `topicListScrollPosition` in that case.

Ahmedgagan:
> Another conflict here ^^
> 
> Also, just checking, does this code run when you click "back" in the browser? We don't want to reset the `topicListScrollPosition` in that case.

Yaa in that case code works fine😃

PR #92, opened 2020-08-04
Refresh members

Ahmedgagan:
Scroll-top on username-filter change


assets/javascripts/discourse-assign/controllers/group-assigned-show.js.es6
```diff
@@ -3,6 +3,7 @@ import UserTopicsList from "discourse/controllers/user-topics-list";
 export default UserTopicsList.extend({
   user: Ember.inject.controller(),
   taskActions: Ember.inject.service(),
+  groupController: Ember.inject.controller('group.assigned'),
```
davidtaylorhq, line 6:
```suggestion
```

assets/javascripts/discourse-assign/controllers/group-assigned-show.js.es6
```diff
@@ -11,7 +12,10 @@ export default UserTopicsList.extend({
     unassign(topic) {
       this.taskActions
         .unassign(topic.get("id"))
-        .then(() => this.send("changeAssigned"));
+        .then(() => {
+          this.send("changeAssigned");
+          this.get('groupController').refreshList();
+        });
```
davidtaylorhq, line 18:
```suggestion
        .then(() => this.send("changeAssigned"));
```

assets/javascripts/discourse-assign/controllers/group-assigned.js.es6
```diff
@@ -29,6 +29,10 @@ export default Controller.extend({
       );
     }
   },
+  refreshList() {
+    console.log("called");
+    this.send("changeAssigned");
+  },
```
davidtaylorhq, line 35:
```suggestion
```

assets/javascripts/discourse-assign/routes/group-assigned.js.es6
```diff
@@ -22,5 +22,12 @@ export default DiscourseRoute.extend({
     } else {
       this.transitionTo("group.assigned.show", "everyone");
     }
+  },
+
+  actions: {
+    changeAssigned() {
+      console.log("route action");
+      this.refresh();
+    }
```
davidtaylorhq, line 31: This bit looks good. The only problem is that we already have a `changeAssigned` action in the `group-assigned-show` route. If you remove that, then I think the action will bubble all the way up to this `group-assigned` route, and the members will get refreshed correctly. (and I think the topic list will also be refreshed, since calling refresh() on `group-assigned` will also refresh child routes)

Ahmedgagan: 😍😍 It worked perfectly fine.
But now the issue is total counts doesn't change after refreshing too

davidtaylorhq:
@Ahmedgagan looks like a JS linting fail in travis

PR #95, opened 2020-08-06
Search

Ahmedgagan:
Basic search added


assets/javascripts/discourse/templates/group/assigned.hbs
```diff
@@ -1,5 +1,6 @@
 <section class="user-secondary-navigation group-assignments">
   {{#mobile-nav class="activity-nav" desktopClass="action-list activity-list nav-stacked" currentPath=router._router.currentPath}}
+    {{input type="text" placeholder="Name/Username" value=filterName class="search"}}
```
davidtaylorhq, line 3: The placeholder should be localised with I18n

app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -144,6 +144,10 @@ def group_members
         .order('COUNT(users.id) DESC')
         .group('users.id')
         .select('users.*, COUNT(users.id) as "assignments_count"')
+
+      members = members.where("users.name like :filter OR users.username like :filter", filter: "%#{params[:filter]}%") if params[:filter]
```
davidtaylorhq, line 148: Interpolating things into SQL strings like this is potentially risky (because someone could inject malicious SQL). I think we should be able to do `members.filter_by_username()` in the same way that the [groups members list does it](https://github.com/discourse/discourse/blob/master/app/controllers/groups_controller.rb#L231)

app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -145,7 +145,7 @@ def group_members
         .group('users.id')
         .select('users.*, COUNT(users.id) as "assignments_count"')
 
-      members = members.where("users.name like :filter OR users.username like :filter", filter: "%#{params[:filter]}%") if params[:filter]
+      members = members.where("users.name ilike ? OR users.username_lower ilike ?", "%#{params[:filter]}%", "%#{params[:filter]}%") if params[:filter]
```
davidtaylorhq, line 148: To avoid repeating the same argument twice, we can use keyword arguments like

```suggestion
      members = members.where("users.name ILIKE :pattern OR users.username_lower ILIKE :pattern", pattern: "%#{params[:filter]}%") if params[:filter]
```

Ahmedgagan: `members = members.where("users.name like :filter OR users.username like :filter", filter: "%#{params[:filter]}%") if params[:filter]` i used this as you specified above that SQL injection can be done in this😅

davidtaylorhq: 🤦  I am so sorry, I read it wrong before. Your approach was 100% fine.

I thought you had done something like

```
.where("users.name LIKE #{filter}")
```

Ahmedgagan: 😂😂  cool ill add a commit right now

PR #97, opened 2020-08-11
Topic search

Ahmedgagan:


plugin.rb
```diff
@@ -221,11 +221,22 @@ class ::ListController
 
     list = apply_ordering(list, options)
 
+    list = list.merge(secure)
+
+    if options[:q].present?
+      term = options[:q]
+      bookmark_ts_query = Search.ts_query(term: term)
```
davidtaylorhq, line 228: Let's rename this to something else ;)

plugin.rb
```diff
@@ -262,11 +274,22 @@ class ::ListController
...
```
davidtaylorhq, line 281: As above

spec/requests/list_controller_spec.rb
```diff
@@ -5,7 +5,10 @@
 
 describe ListController do
 
-  before { SiteSetting.assign_enabled = true }
+  before do
+    SiteSetting.assign_enabled = true
+    SearchIndexer.enable
```
davidtaylorhq, line 10: The reason that we disable search indexing by default is that it is quite slow. Can you move this so that it is only enabled in the `filtering of topics` context? That way, it won't make the other tests slower

Ahmedgagan: Sure ill do that right now😊

PR #98, opened 2020-08-13
Remember sort and topic-search

Ahmedgagan:


assets/javascripts/discourse/templates/group/assigned.hbs
```diff
@@ -6,9 +6,9 @@
       </div>
     {{/if}}
     {{#load-more selector=".activity-nav li" action=(action "loadMore")}}
-      {{group-assigned-filter show-avatar=false filter="everyone" routeType=route_type assignment_count=group.assignment_count}}
+      {{group-assigned-filter show-avatar=false filter="everyone" routeType=route_type assignment_count=group.assignment_count router=router}}
```
davidtaylorhq, line 9: Ideally components should be separated from the application, so it would be best if we didn't pass them the router. Can you pass the necessary variables like `q=q ascending=ascending order=order` instead?

Ahmedgagan: Sure, ill make this change

Ahmedgagan: Done in the latest commit 😃

assets/javascripts/discourse-assign/controllers/group-assigned.js.es6
```diff
@@ -12,6 +12,30 @@ export default Controller.extend({
   filterName: "",
   filter: "",
 
+  @discourseComputed("router.currentRoute.queryParams.order")
```
davidtaylorhq, line 15: I guess this needs to be changed since the router is gone?

Ahmedgagan: we are getting router in the controller [here](https://github.com/discourse/discourse-assign/blob/83ff23c98ec8d5edcb3090093c4025db1d618ef0/assets/javascripts/discourse-assign/controllers/group-assigned.js.es6#L8) we still need to change this?

davidtaylorhq: Hmm I see... It is much better now that it's in the controller, but still, injecting the router service is not ideal.

What happens if we add order/ascending/q as queryParams to this controller? Does it break the group-assigned-show controller? Or does it work ok?

Ahmedgagan: It breaks the `group-assigned-show` controller

davidtaylorhq: Ok, let's keep this then :+1:

assets/javascripts/discourse-assign/controllers/group-assigned.js.es6
```diff
@@ -12,6 +12,30 @@ export default Controller.extend({
...
+  order(order) {
+    if (order) {
+      return order;
+    }
+    return "";
```
davidtaylorhq, line 20: How about 

```suggestion
    return order || "";
```

Ahmedgagan: Updated in the latest commit 😃

davidtaylorhq:
@Ahmedgagan please can you resolve the conflict

Ahmedgagan:
> @Ahmedgagan please can you resolve the conflict

Solved in latest commit

assets/javascripts/discourse/templates/group/assigned.hbs
```diff
@@ -12,9 +12,23 @@
...
+        q=q
+        ascending=ascending
+        order=order}}
       {{#each members as |member|}}
-        {{group-assigned-filter show-avatar=true filter=member routeType=route_type}}
+        {{group-assigned-filter
+          show-avatar=true
+          filter=member
+          routeType=route_type
+          assignment_count=group.assignment_count
```
davidtaylorhq, line 28: Do we need to pass the group assignment count here? I think we only need it for the 'everyone' button?

assets/javascripts/discourse/templates/group/assigned.hbs
```diff
@@ -12,9 +12,23 @@
       </div>
     {{/if}}
     {{#load-more selector=".activity-nav li" action=(action "loadMore")}}
-      {{group-assigned-filter show-avatar=false filter="everyone" routeType=route_type assignmentCount=group.assignment_count}}
+      {{group-assigned-filter
+        show-avatar=false
+        filter="everyone"
+        routeType=route_type
+        assignment_count=group.assignment_count
```
davidtaylorhq, line 19: Does this work? I think we just changed this to `assignmentCount` rather than `assignment_count`

Ahmedgagan: Yaa you are right😅 I just added a commit changing this

PR #99, opened 2020-08-13
DEV: various code comments

jjaffeux:
@davidtaylorhq @Ahmedgagan this is various code comments on this file, that I hope you will find useful. Feel free to scrap it, it's not meant to be merged.

### DIFF
--- a/assets/javascripts/discourse-assign/controllers/group-assigned.js.es6
+++ b/assets/javascripts/discourse-assign/controllers/group-assigned.js.es6
@@ -1,8 +1,10 @@
 import { inject as service } from "@ember/service";
 import Controller, { inject as controller } from "@ember/controller";
+import { action } from "@ember/object";
+import { debounce } from "@ember/runloop";
 import { ajax } from "discourse/lib/ajax";
 import discourseComputed, { observes } from "discourse-common/utils/decorators";
-import discourseDebounce from "discourse/lib/debounce";
+import { INPUT_DELAY } from "discourse-common/config/environment";
 
 export default Controller.extend({
   router: service(),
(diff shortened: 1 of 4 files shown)


assets/javascripts/discourse-assign/controllers/group-assigned.js.es6
```diff
@@ -12,35 +14,62 @@ export default Controller.extend({
...
-      }
-      this.set("members", result.members);
-      this.set("loading", false);
-    });
-  },
+    })
+      .then(result => {
+        // know that this is risky, and a bug actually in your code
+        // if you type something in filter and click fast on logo befire the end of debouncing
+        // you will load home and then reload group.assigned.show
```
davidtaylorhq, line 61: How would you suggest we improve this @jjaffeux? I guess we can check something like `router.isActive('group-assigned-show')` first?

assets/javascripts/discourse-assign/controllers/group-assigned.js.es6
```diff
@@ -12,35 +14,62 @@ export default Controller.extend({
   filterName: "",
   filter: "",
 
+  // I don't know know the subject enough but we should try to remove this one too
+  // one way to do this would be to maybe pass group to group-assigned-filter
+  // and use didReceiveAttrs, I didn't try, but that might be a path forward
```
davidtaylorhq, line 19: This is solving the problem

- We have the `group` route, which loads the group info, including the assigned count. We use that to put a counter like <kbd>Assigned (3)</kbd> on the tab

- When you're in the `group-assigned` tab, it is possible to unassign topics. Therefore the count in the tab should change

Reloading the entire `group` model seems wasteful, so we're just updating the assignment count when needed.

We can totally move this to an `afterModel` hook on the `group-assigned` route. But I still don't really like how we are modifying data in a parent route 🤔  It violates the DDAU principle.

Maybe `group-assigned` can fire an action `assignedCountChanged(count)` during `afterModel`, and then `group` can implement that action, and update the assignment count.

assets/javascripts/discourse-assign/controllers/group-assigned.js.es6
```diff
@@ -12,35 +14,62 @@ export default Controller.extend({
...
-      this.set("members", result.members);
-      this.set("loading", false);
-    });
-  },
+    })
+      .then(result => {
+        // know that this is risky, and a bug actually in your code
+        // if you type something in filter and click fast on logo befire the end of debouncing
+        // you will load home and then reload group.assigned.show
+        if (this.router.currentRoute.params.filter !== "everyone") {
```
davidtaylorhq, line 62: @jjaffeux how do you feel about accessing the route service like this in a controller? Ahmed and I have been talking in [this PR](https://github.com/discourse/discourse-assign/pull/98#discussion_r470098636) about it... 

Originally I was thinking it should be avoided... but now I'm reading [the ember docs](https://api.emberjs.com/ember/3.9/classes/RouterService/methods) about the service and it does seem to be designed for use by controllers and components. So maybe this is fine? 🤷‍♂️

jjaffeux: I would yes it's fine, and you are only accessibg public things so 🤷‍♂️

assets/javascripts/discourse-assign/controllers/group-assigned.js.es6
jjaffeux: Yes seems like a good path

assets/javascripts/discourse-assign/controllers/group-assigned.js.es6
jjaffeux: I might be missing something, but I think https://github.com/discourse/discourse-assign/pull/99/commits/3a77beef427670c50cc309db1d41ca766e02e925 should just work ?

Ahmedgagan: When any unassign happens, we need to refresh the `group.assignment_count` which is available in the parent route. we need to update `group.assignment_count` as we use it in the parent route too. so I think this might not work here?

jjaffeux: Please do me a test case where it doesn't work if possible

Ahmedgagan: [here](https://github.com/discourse/discourse-assign/blob/83ff23c98ec8d5edcb3090093c4025db1d618ef0/assets/javascripts/discourse/templates/components/group-assigned-menu-item.hbs#L2) we use `assignment_count` from group model. so, whenever a reassign happens we need to update assignment_count from group model

jjaffeux: ok indeed I missed the connector, in this case I think just doing this: https://github.com/discourse/discourse-assign/pull/99/commits/725a79eb93268a7b5a1a3d855204d66bcff8bb7b

would be better, it's a little bit hacky but okayish IMO, and still better than an observer

Ahmedgagan: Screenshot 2020-08-14 at 3 05 28 AM
This happens when we unassign any topic, tab count remains same and just side-bar counts changes

Ahmedgagan: That's Perfect😍

davidtaylorhq:
Comments integrated in 99c2c48c. Thanks @jjaffeux

PR #102, opened 2020-08-15
Advanced search

Ahmedgagan:
`in:assigned` and `in:not_assigned` filter's added in advanced search options


plugin.rb
```diff
@@ -497,4 +497,29 @@ def self.enqueue_assign_hooks(event, payload)
     end
   end
 
+  require_dependency 'search'
+
+  #TODO Remove when plugin is 1.0
+  if Search.respond_to? :advanced_filter
+    Search.advanced_filter(/in:assigned/) do |posts|
```
davidtaylorhq, line 504: I think you should be able to replace all this with the [plugin API](https://github.com/discourse/discourse/blob/master/lib/plugin/instance.rb#L205-L205)
```suggestion
register_search_advanced_filter(/in:assigned/) do |posts|
```

assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
```diff
@@ -319,6 +320,22 @@ export default {
       return;
     }
 
+    SearchAdvancedOptions.reopen({
+      didInsertElement() {
```
davidtaylorhq, line 324: These arrays of options are set up in `init()`. So I think it might be cleaner to override `init()` rather than `didInsertElement`

Ahmedgagan:
> Approach is looking good to me. We will need some JS and Ruby tests for this. And we also need to make sure this is secure behind the `guardian.can_assign?` check.

Where can I add the check `guardian.can_assign?` as we are working with the instance.rb method where `guardian` and `current_user` both are not available

davidtaylorhq:
Does `@guardian` work? I think that's how other advanced search filters work

assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
```diff
@@ -318,6 +320,57 @@ export default {
     if (!siteSettings.assign_enabled) {
       return;
     }
+    const currentUser = container.lookup("current-user:main");
+    if (currentUser.can_assign) {
+      const REGEXP_USERNAME_PREFIX = /^(user:|@)/gi;
```
davidtaylorhq, line 325: This filter seems copied from the normal user selector (where you can use `@david` or `user:david`). Do we need it here?

plugin.rb
```diff
@@ -497,4 +497,40 @@ def self.enqueue_assign_hooks(event, payload)
     end
   end
 
+  register_search_advanced_filter(/in:assigned/) do |posts|
+    raise Discourse::InvalidAccess unless @guardian.can_assign?
```
davidtaylorhq, line 501: How does this look from the UI? I think it might be better to make this search do nothing, rather than raise an error. What do you think?

assets/javascripts/discourse-assign/connectors/advanced-search-options-above/assigned-advanced-search.hbs
```diff
@@ -0,0 +1,12 @@
+<div class="control-group pull-left">
+  <label class="control-label" for="search-posted-by">{{i18n "search.advanced.assigned.label"}}</label>
```
davidtaylorhq, line 2: I think `search-posted-by` probably needs to be updated?

plugin.rb
Ahmedgagan: Yaa that's right.

assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
Ahmedgagan: yaa ill change the regex

plugin.rb
```diff
@@ -497,4 +497,40 @@ def self.enqueue_assign_hooks(event, payload)
...
+        FROM topic_custom_fields tc
+        WHERE tc.name = 'assigned_to_id' AND
+                        tc.value IS NOT NULL
+        )")
+    end
+  end
+
+  register_search_advanced_filter(/assigned:(.+)$/) do |posts, match|
+    if @guardian.can_assign?
+      user = User.find_by_username(match)
```
ZogStriP, line 524: What happens if `user` can't be found (eg. the username is wrong)?

assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
```diff
@@ -318,6 +320,59 @@ export default {
     if (!siteSettings.assign_enabled) {
       return;
     }
+    const currentUser = container.lookup("current-user:main");
+    if (currentUser.can_assign) {
+      const REGEXP_USERNAME_PREFIX = /^(assigned:)/gi;
```
ZogStriP, line 325: This `const` should probably be extracted out of that method.

plugin.rb
Ahmedgagan: its a user-picker, so if the username is wrong it won't return any user options. I've sent you a screen-recording on meta do have a look😃

ZogStriP: Sure, but what about if someone manually forges a request with a wrong username?

Ahmedgagan: There's an error in the backend but the front end handles it 
Screenshot 2020-08-14 at 5 01 36 PM

ZogStriP: Would be even better if there was no error in the backend 😉

assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
```diff
@@ -312,13 +314,67 @@ function initialize(api) {
   api.addKeyboardShortcut("g a", "", { path: "/my/activity/assigned" });
 }
 
+export const REGEXP_USERNAME_PREFIX = /^(assigned:)/gi;
```
ZogStriP, line 317: Is this being used anywhere else? If not, then you don't need to `export` it 😉

assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
```diff
@@ -312,13 +314,67 @@ function initialize(api) {
...
   name: "extend-for-assign",
   initialize(container) {
     const siteSettings = container.lookup("site-settings:main");
     if (!siteSettings.assign_enabled) {
       return;
     }
+    const currentUser = container.lookup("current-user:main");
+    if (currentUser.can_assign) {
+      SearchAdvancedOptions.reopen({
+        _init() {
```
ZogStriP, line 329: Instead of overwriting the `SearchAdvancedOptions._init()` method, we should add a new "plugin API" in core to allow plugin to add new "in options".

plugin.rb
```diff
@@ -521,4 +521,43 @@ def self.enqueue_assign_hooks(event, payload)
...
+        FROM topic_custom_fields tc
+        WHERE tc.name = 'assigned_to_id' AND
+                        tc.value IS NOT NULL
+        )")
+    end
+  end
+
+  register_search_advanced_filter(/assigned:(.+)$/) do |posts, match|
+    if @guardian.can_assign?
+      user_id = User.where(staged: false).where('username_lower = ? OR id = ?', match.downcase, match.to_i).pluck_first(:id)
```
ZogStriP, line 548: I liked `User.find_by_username` better so we don't duplicate the logic. I was thinking something like this

```ruby
register_search_advanced_filter(/assigned:(.+)$/) do |posts, match|
  if @guardian.can_assign?
    if user_id = User.find_by_username(match)&.id
      post.where("..... #{user_id}")
    end
  end
end
```

That way it will only filter when the username matches a real user and won't generate an error when it doesn't 😉

Ahmedgagan:
`Instead of overwriting the SearchAdvancedOptions._init() method, we should add a new "plugin API" in core to allow plugin to add new "in options".`
@ZogStriP I'll try to add it and make a PR in the core for this.

assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
```diff
@@ -312,16 +326,70 @@ function initialize(api) {
...
+          // this.inOptionsForUsers.pushObjects([
+          //   {
+          //     name: I18n.t("search.advanced.in.assigned"),
+          //     value: "assigned"
+          //   },
+          //   {
+          //     name: I18n.t("search.advanced.in.not_assigned"),
+          //     value: "not_assigned"
+          //   }
+          // ]);
```
davidtaylorhq, line 351: Can we 🗑️  this commented code?

plugin.rb
```diff
@@ -521,4 +521,40 @@ def self.enqueue_assign_hooks(event, payload)
...
+      posts.where("topics.id IN (
+        SELECT tc.topic_id
+        FROM topic_custom_fields tc
+        WHERE tc.name = 'assigned_to_id' AND
+                        tc.value IS NOT NULL
+        )")
+    end
+  end
+
+  register_search_advanced_filter(/in:not_assigned/) do |posts|
```
davidtaylorhq, line 535: Can we make this `in:unassigned`. Then it will match the rest of the UI better:

Screenshot 2020-09-09 at 15 47 40


(Sorry, I know this means changing it in a few places 😓 )

davidtaylorhq:
Looks like travis is failing on a prettier issue. We recently updated the prettier version in core, so you may need to run `yarn install` in your discourse directory, then re-format the `extend-for-assigns.js.es6` file.

A couple of other comments above, but this is looking really nice now that we have the core API 😁

assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
```diff
@@ -312,16 +326,70 @@ function initialize(api) {
...
+            REGEXP_USERNAME_PREFIX
+          );
+        },
+      });
+    }
 
-    withPluginApi("0.8.11", (api) => initialize(api, container));
-    withPluginApi("0.8.28", (api) =>
+    withPluginApi("0.11.0", (api) => initialize(api, container));
+    withPluginApi("0.11.0", (api) =>
```
davidtaylorhq, line 392: Changing this will stop the plugin working on Discourse Stable. We only need the new API for the advanced search changes in `initialize()`. `registerTopicFooterButtons` can stay on 0.8.28

```suggestion
    withPluginApi("0.8.28", (api) =>
```

test/javascripts/acceptance/search-full-test.js.es6
```diff
@@ -0,0 +1,91 @@
...
+    );
+    assert.equal(
+      find(".search-query").val(),
+      "none in:not_assigned",
+      'has updated search term to "none in:not_assinged"'
+    );
+  }
+);
+
+QUnit.skip("update assigned to through advanced search ui", async (assert) => {
```
davidtaylorhq, line 56: I don't think this should be `.skip`?

assets/javascripts/discourse-assign/connectors/advanced-search-options-above/assigned-advanced-search.hbs
```diff
@@ -0,0 +1,12 @@
+<div class="control-group pull-left">
```
davidtaylorhq, line 1: Can we move this to the advanced-search-options**-below** outlet? I don't think it should be right at the top, since it is for a plugin.

Ahmedgagan:
`Does this still need to be .skip?`
Yes, as in the core to it skipped 😅

assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
```diff
@@ -312,15 +327,58 @@ function initialize(api) {
...
+          );
+        },
+      });
+    }
 
-    withPluginApi("0.8.11", (api) => initialize(api, container));
+    withPluginApi("0.11.0", (api) => initialize(api, container));
     withPluginApi("0.8.28", (api) =>
       registerTopicFooterButtons(api, container)
     );
```
ZogStriP, line 384: Why not merge the 2 calls to `withPluginApi`?

```suggestion
    withPluginApi("0.11.0", (api) => { 
      initialize(api, container);
      registerTopicFooterButtons(api, container);
    });
```

davidtaylorhq: We need to use different API versions. The new search dropdown APIs are only in 0.11.0, but we want `registerTopicFooterButtons` to continue working on the stable version of core.

yarn-error.log
```diff
@@ -0,0 +1,1649 @@
+Arguments:
```
davidtaylorhq, line 1: This file needs to be removed

assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
```diff
@@ -119,6 +121,19 @@ function initialize(api) {
...
+        name: I18n.t("search.advanced.in.assigned"),
+        value: "assigned",
+      },
+      {
+        name: I18n.t("search.advanced.in.unassigned"),
+        value: "unassigned",
+      },
+    ],
+  });
+
```
davidtaylorhq, line 136: Do we need to check `currentUser.can_assign` here? We don't want to show this to all users

assets/javascripts/discourse-assign/connectors/advanced-search-options-below/assigned-advanced-search.js.es6
```diff
@@ -0,0 +1,8 @@
+export default {
+  shouldRender(args, component) {
+    return (
+      component.currentUser &&
+      component.currentUser.can_assign &&
```
davidtaylorhq, line 5:
```suggestion
      component.currentUser.can_assign
```

assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
```diff
@@ -119,6 +121,23 @@ function initialize(api) {
     before: "top",
   });
 
+  api.addAdvancedSearchOptions(
+    api.container.lookup("current-user:main").can_assign
```
davidtaylorhq, line 125:
```suggestion
    api.getCurrentUser() && api.getCurrentUser().can_assign
```

PR #103, opened 2020-08-20
Added restrictions to show `assigned tab`

Ahmedgagan:


plugin.rb
```diff
@@ -71,6 +75,16 @@
     @can_assign == :true
   end
 
+  add_to_class(:group, :can_show_assigned_tab?) do
+    users = User.joins("JOIN group_users gu on users.id=gu.user_id").where("gu.group_id=?", self.id)
+
+    users.each do |u|
```
davidtaylorhq, line 81: There might be thousands of users here, so we can't afford to load them into memory and loop over them. We will need to write a single SQL query which does the check.

PR #107, opened 2020-09-01
Feature: Category and tag filter as bread-crumbs for topics filtering in group-assigned page

Ahmedgagan:


assets/javascripts/discourse/components/tag-drop-assigned.js.es6
```diff
@@ -0,0 +1,22 @@
+import TagDrop from "select-kit/components/tag-drop";
+
+export const NO_TAG_ID = "no-tags";
+export const ALL_TAGS_ID = "all-tags";
+export const NONE_TAG_ID = "none";
+
+export default TagDrop.extend({
+  actions: {
+    onChange(tagId) {
+      switch (tagId) {
```
ZogStriP, line 10: I think the `if` / `else if` / `else` approach like you did for the categories is a tiny bit easier on the eyes 😉

Ahmedgagan: Sure, ill make that change 😃

assets/javascripts/discourse/templates/components/bread-crumbs-assigned.hbs
```diff
@@ -0,0 +1,19 @@
+{{#each categoryBreadcrumbs as |breadcrumb|}}
```
ZogStriP, line 1: Why do you need to create this new bread crumb component? Can't you re-use the [one in core](https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/app/templates/components/bread-crumbs.hbs)?

assets/javascripts/discourse/components/category-drop-assigned.js.es6
```diff
@@ -0,0 +1,18 @@
+import CategoryDrop from "select-kit/components/category-drop";
+
+export const NO_CATEGORIES_ID = "no-categories";
```
ZogStriP, line 3: Is this being used anywhere?

assets/javascripts/discourse/components/category-drop-assigned.js.es6
```diff
@@ -0,0 +1,18 @@
...
+export const ALL_CATEGORIES_ID = "all-categories";
```
ZogStriP, line 4: Is this being used anywhere?

assets/javascripts/discourse/components/tag-drop-assigned.js.es6
```diff
@@ -0,0 +1,19 @@
+import TagDrop from "select-kit/components/tag-drop";
+
+export const NO_TAG_ID = "no-tags";
+export const ALL_TAGS_ID = "all-tags";
+export const NONE_TAG_ID = "none";
```
ZogStriP, line 5: Is this being used anywhere?

assets/javascripts/discourse/templates/components/bread-crumbs-assigned.hbs
Ahmedgagan: as we need to add a `categoryId` variable in bread-crumbs-assigned to add in query-params so, I thought this will be the best approach. your thoughts?

assets/javascripts/discourse/components/category-drop-assigned.js.es6
Ahmedgagan: we are using this in `else if` statement

assets/javascripts/discourse/templates/components/bread-crumbs-assigned.hbs
ZogStriP: Do you need the `categoryId` when you can extract the `id` of the `category`?

assets/javascripts/discourse/components/category-drop-assigned.js.es6
ZogStriP: There's no need to export it then. Also, looks like it's already defined [in core](https://github.com/discourse/discourse/blob/c200238bdc241586bfe82b2d0fc1f600fec45b1f/app/assets/javascripts/select-kit/addon/components/category-drop.js#L10-L11) so you might want to import it instead ;)

assets/javascripts/discourse/templates/components/bread-crumbs-assigned.hbs
Ahmedgagan: That's right but we need to override `onChange` in category-drop, so may I keep that empty? as in `category-drop`, `onChange` method redirects to URL of that category

assets/javascripts/discourse/components/category-drop-assigned.js.es6
Ahmedgagan: Right, ill add an import statement for this😃

assets/javascripts/discourse/templates/components/bread-crumbs-assigned.hbs
Ahmedgagan: I think we need to do that, as we need to override category-drop and to override that too we need to override `bread-crumbs` template

plugin.rb
```diff
@@ -246,6 +246,50 @@ class ::ListController
 
     list = list.merge(secure)
 
+    if SiteSetting.tagging_enabled
```
ZogStriP, line 249: Looks like this is a copy & paste from [`topic_query`](https://github.com/discourse/discourse/blob/de038c0eab576317fe7cbb647e4388fad7cc566e/lib/topic_query.rb#L693-L735), any chances you could instead use `TopicQuery`?

ZogStriP: Have you tried doing something like this?

```ruby
add_to_class(:topic_query, :list_messages_assigned) do |user|
  create_list(:assigned, unordered: true) do |topics|
    # here are the specific filters for the assigned topics list
    topics.where(...)
  end
end
```

Ahmedgagan: Yaa that's right but it also uses the `default_results` method to do all the filtering work. so if we re-use this method it will load all the topics and then filter assigned ones, it won't be feasible than right?

ZogStriP: > so if we re-use this method it will load all the topics and then filter assigned ones

Not exactly, the `default_results` method does not load all the topics. It returns an ActiveRecord object that can then be used to add more "filter" to the query before it's being executed.

Ahmedgagan: Oooh, then it will make our task much easier😅, I'll refector the code accordingly in next commit.....thanks a lot 😅

Ahmedgagan: Hey, @ZogStriP I tried this out, but this doesn't include PM, so i think we can't use this method

davidtaylorhq: Yeah you're right, `default_results` doesn't work for PMs. But now this plugin is copying so much from core... that maybe we need to try and make some changes in core to make it work better. I will take a look and see if there are any easy changes which can help us.

We have a similar problem (needing topics + PMs in the same list) for some places in core as well (e.g. bookmarks, drafts), so I think a refactor might be needed.

davidtaylorhq:
@Ahmedgagan I tried this out locally but it doesn't seem to be working quite right. When I visit the group assigned page, all categorized topics are missing from the list. I only see uncategorized topics and PMs. The URL in my address bar is

```
http://localhost:3000/g/trust_level_4/assigned/everyone
```

In dev tools I can see it is loading data from

```
http://localhost:3000/topics/group-topics-assigned/trust_level_4.json?order=&ascending=false&category=&tags=&no_tags=null&search=
```
So maybe because it is sending `category=` without a value, it is only looking for topics with no category?

Ahmedgagan:
> @Ahmedgagan I tried this out locally but it doesn't seem to be working quite right. When I visit the group assigned page, all categorized topics are missing from the list. I only see uncategorized topics and PMs. The URL in my address bar is
> 
> ```
> http://localhost:3000/g/trust_level_4/assigned/everyone
> ```
> 
> In dev tools I can see it is loading data from
> 
> ```
> http://localhost:3000/topics/group-topics-assigned/trust_level_4.json?order=&ascending=false&category=&tags=&no_tags=null&search=
> ```
> 
> So maybe because it is sending `category=` without a value, it is only looking for topics with no category?

Sure, ill take a look 😃

Ahmedgagan:
Hi @davidtaylorhq can you please have a look at it again whenever you get time? It's been long 😀

davidtaylorhq:
It seems to be working better with categories now, but I think there is the same issue with tags. If I have 'all tags' selected, then it only showing 'untagged' topics. I have to select the tag before they appear.

davidtaylorhq:
Thanks for the commits @Ahmedgagan, I will take another look and hopefully merge later this week

Ahmedgagan:
There's a small thing left in this, adding a commit in a bit

CvX:
@Ahmedgagan @davidtaylorhq should this PR be updated or closed?

Ahmedgagan:
> @Ahmedgagan @davidtaylorhq should this PR be updated or closed?

If the feature seems acceptable to you guys, i can update the PR.

davidtaylorhq:
I think there would be quite a few changes required to make this ready for merging (rebasing onto latest `main`, adding JS tests, etc.). So I think we should close for now - we haven't had any requests for this feature, and I think the API gem work is higher priority. We can always re-open in future.

PR #110, opened 2020-09-09
FEATURE: Bulk Actions on Topics List

Ahmedgagan:


davidtaylorhq:
Just checked this out locally and it feels very nice to use 😄  A few comments, some of which you probably already know:

- This will need to be updated once we have the new core API to permit parameters.

- Let's update the button text to "Assign" and "Unassign"

- Can we add icons to match the normal assign/unassign buttons?

- We need some tests here, especially on the server-side

PR #111, opened 2020-09-10
REFACTOR: Use core `default_results` function

davidtaylorhq:
This significantly reduces the amount of logic we need to carry in the discourse-assign plugin. One side effect is that I had to rename the 'q' parameter to 'search', so that it matches core's implementation.

Requires https://github.com/discourse/discourse/pull/10647 to be merged first


eviltrout:
So this is a breaking change to anyone who has an API set up to use `q`?

davidtaylorhq:
Technically yes, although we only added the q parameter to assigned topic lists a couple of weeks ago, so i would be surprised if many people are using it via the API.

eviltrout:
@davidtaylorhq if that's the case then I have no problems changing it without deprecations.

eviltrout (approved):

PR #113, opened 2020-09-24
FIX: Advanced-search-option for assigned-user

Ahmedgagan:


eviltrout:
This PR seems to have no changes? I think it might have been made incorrectly.

Ahmedgagan:
Oooh I think there's some issue in committing the code😅 I'll add the commit in this PR

jjaffeux:
Yes I hesitated to do this, I'm afraid it will break old Discourse installs.

davidtaylorhq:
@jjaffeux discourse-assign is currently pinned for all discourse versions before 2.6.0.beta1 https://github.com/discourse/discourse-assign/blob/master/.discourse-compatibility#L1

Do you think it would be ok with beta1 and above? Or should we add a pin for beta2 in this PR?

jjaffeux:
Yes I think it should be beta 2, this commit will work from beta 3 for sure (and slightly before too but... not much se can doa bout this).

SamSaffron:
depends on how this breaks, I guess... if it explodes Discourse on beta 2 then we should not merge. But if only some functionality in advanced search is a bit wonky I guess we live with this. @jjaffeux  can you clarify ?

CvX:
@Ahmedgagan @jjaffeux should this PR be updated or closed?

Ahmedgagan:
If the feature seems acceptable to you guys, i can update the PR.

Ahmedgagan:
> @Ahmedgagan @jjaffeux should this PR be updated or closed?

If the feature seems acceptable to you guys, i can update the PR.

davidtaylorhq:
I think @OsamaSayegh took care of fixing this bug during refactoring in https://github.com/discourse/discourse-assign/commit/b1b8c94f566b298d0d0640e12b0eec2d26defdef. As far as I can tell, advanced search UI is working well on Meta. 😄

PR #114, opened 2020-09-30
FIX: removes the need for intermediate property and updates code

jjaffeux:
Before this commit, changing the value of the reminder frequency on a user notifications page was not working.

This commit also get rid of a few deprecations warning.

### DIFF
--- a/assets/javascripts/discourse/components/remind-assigns-frequency.js.es6
+++ b/assets/javascripts/discourse/components/remind-assigns-frequency.js.es6
@@ -1,10 +1,19 @@
 import I18n from "I18n";
-import computed from "discourse-common/utils/decorators";
+import { computed } from "@ember/object";
+import discourseComputed from "discourse-common/utils/decorators";
 
 export default Ember.Component.extend({
-  selectedFrequency: null,
+  selectedFrequency: computed(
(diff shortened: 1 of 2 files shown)


jjaffeux:
Screenshot 2020-09-30 at 10 36 15

assets/javascripts/discourse/components/remind-assigns-frequency.js.es6
```diff
@@ -1,10 +1,19 @@
 import I18n from "I18n";
-import computed from "discourse-common/utils/decorators";
+import { computed } from "@ember/object";
+import discourseComputed from "discourse-common/utils/decorators";
 
 export default Ember.Component.extend({
-  selectedFrequency: null,
+  selectedFrequency: computed(
```
CvX, line 6: That looks like a good candidate for the `@or` computed helper?

jjaffeux: Yes it crossed my mind while reading the code and thinking what I would change and forgot after :p will change :+1:

jjaffeux: done

CvX (approved):

PR #121, opened 2020-12-08
FEATURE: show assigned user on search results

techAPJ:
Related PR on core: https://github.com/discourse/discourse/pull/11428

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -346,6 +346,20 @@ after_initialize do
     end
   end
 
+  add_to_class(:search_topic_list_item_serializer, :assigned_to_user_id) do
+    id = object.custom_fields[TopicAssigner::ASSIGNED_TO_ID]
+    # a bit messy but race conditions can give us an array here, avoid
+    id && id.to_i rescue nil
+  end
+
+  add_to_serializer(:search_topic_list_item, :assigned_to_user, false) do
+    object.assigned_to_user
+  end
+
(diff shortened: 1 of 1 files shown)


plugin.rb
```diff
@@ -346,6 +346,19 @@ class ::ListController
     end
   end
 
+  add_to_class(:search_topic_list_item_serializer, :assigned_to_user_id) do
+    id = object.custom_fields[TopicAssigner::ASSIGNED_TO_ID]
+    id && id.to_i rescue nil
```
ZogStriP, line 351: I don't think `to_i` ever raise an exception. `nil.to_i == 0`

techAPJ: Looking at code comment:

https://github.com/discourse/discourse-assign/blob/c4aefd5917be8f5b897cfede19fd120deb3c8a6c/plugin.rb#L394-L395

I think it's for array (`[].to_i`) cases.

ZogStriP: I see, I guess it might be worth adding the same comment?

techAPJ: Makes sense, done.

ZogStriP (approved):

plugin.rb
tgxworld: I highly recommend never using `rescue nil` because it rescues all and any type of errors.

eviltrout: I've seen you recommend this before (and correctly.) I wonder if we can lint for it?

PR #123, opened 2020-12-17
FEATURE: show assigned user on search results

techAPJ:
This commit allows discourse-assign plugin to show assigned users on search result topic list.

Accompanying PR on meta: https://github.com/discourse/discourse/pull/11518

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -129,6 +129,7 @@ after_initialize do
 
   TopicList.preloaded_custom_fields << TopicAssigner::ASSIGNED_TO_ID
   Site.preloaded_category_custom_fields << "enable_unassigned_filter" if Site.respond_to? :preloaded_category_custom_fields
+  Search.preloaded_topic_custom_fields << TopicAssigner::ASSIGNED_TO_ID if Search.respond_to? :preloaded_topic_custom_fields
 
   if defined? BookmarkQuery
     if BookmarkQuery.respond_to?(:preloaded_custom_fields) && BookmarkQuery.respond_to?(:on_preload)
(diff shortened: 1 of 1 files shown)


plugin.rb
```diff
@@ -182,6 +183,28 @@
     end
   end
 
+  if Search.respond_to?(:on_preload)
+    Search.on_preload do |search, search_query|
```
CvX, line 187: Mirroring the comment from the other PR, are `search` and `search_query` the best names for those variables?

plugin.rb
```diff
@@ -182,6 +183,28 @@
...
+      if SiteSetting.assign_enabled?
+        can_assign = search_query.guardian && search_query.guardian.can_assign?
+        allowed_access = SiteSetting.assigns_public || can_assign
+
+        if allowed_access && search.posts.length > 0
+          assigned_user_ids = search.posts.map(&:topic).map { |topic| topic.custom_fields[TopicAssigner::ASSIGNED_TO_ID] }.compact.uniq
+          assigned_users = {}
+          User.where(id: assigned_user_ids).each do |user|
+            assigned_users[user.id] = user
+          end
```
CvX, line 197:
```suggestion
          assigned_users = User.where(id: assigned_user_ids).index_by(&:id)
```

CvX (changes requested):

plugin.rb
techAPJ: Makes sense, updated variable names.

plugin.rb
techAPJ: Thanks for the tip! 👍

CvX (approved):

PR #124, opened 2020-12-22
DEV: Use the new discourseDebounce function wrapper.

romanrizzi:
We recently merged a Discourse core's PR to replace usages of Ember's debounce and discourseDebounce with a new debounce wrapper. The new wrapper works exactly like Ember's debounce but internally calls "run" when called in test mode.

This PR replaces all usages of other debounce functions with the new wrapper and fallbacks to Ember's debounce for backward-compatibility.

### DIFF
--- a/assets/javascripts/discourse-assign/controllers/group-assigned-show.js.es6
+++ b/assets/javascripts/discourse-assign/controllers/group-assigned-show.js.es6
@@ -1,7 +1,7 @@
 import UserTopicsList from "discourse/controllers/user-topics-list";
 import { alias } from "@ember/object/computed";
-import { debounce } from "@ember/runloop";
 import discourseComputed from "discourse-common/utils/decorators";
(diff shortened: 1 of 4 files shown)


assets/javascripts/lib/get-debounce.js.es6
```diff
@@ -0,0 +1,11 @@
...
+import { debounce } from "@ember/runloop";
+
+// TODO: Remove this file and use discouseDebounce after the 2.7 release.
+export default function () {
+  if (discourseDebounce === undefined) {
+    return debounce;
+  }
+
+  return discourseDebounce;
+}
```
CvX, line 11: How about renaming this file to `debounce.js.es6` and exporting a function instead of a function factory?

```suggestion
const debounceFunction = discourseDebounce || debounce;
export default debounceFunction;
```

All call sites could then remain unchanged.

CvX (approved):

eviltrout (approved):

PR #147, opened 2021-02-22
 FIX: Use Ember's debounce on stable.

romanrizzi:
We need to wrap the new debounce function inside a try block to avoid throwing a "module not found" exception.

### DIFF
--- a/assets/javascripts/lib/debounce.js.es6
+++ b/assets/javascripts/lib/debounce.js.es6
@@ -1,6 +1,10 @@
-import discourseDebounce from "discourse-common/lib/debounce";
 import { debounce } from "@ember/runloop";
 
 // TODO: Remove this file and use discouseDebounce after the 2.7 release.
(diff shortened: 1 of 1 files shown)


gschlager (approved):

assets/javascripts/lib/debounce.js.es6
```diff
@@ -1,6 +1,10 @@
-import discourseDebounce from "discourse-common/lib/debounce";
 import { debounce } from "@ember/runloop";
 
 // TODO: Remove this file and use discouseDebounce after the 2.7 release.
-const debounceFunction = discourseDebounce || debounce;
+let debounceFunction = debounce;
+
+try {
```
eviltrout, line 6: FYI you can check if a module exists without a try/catch by looking at `require._eak_seen`. Catching exceptions is quite slow.

In this case it's probably not a problem because it's only run once

PR #162, opened 2021-06-24
FEATURE: adds support for a random assign automation

jjaffeux:


plugin.rb
```diff
@@ -575,4 +575,32 @@ def self.enqueue_assign_hooks(event, payload)
...
+      field :assignees_group, component: :group
+      field :assigned_topic, component: :text
+
+      version 1
+
+      triggerables %i[point_in_time recurring]
+
+      script do |context, fields|
+        begin
+          next unless SiteSetting.assign_enabled
```
eviltrout: If you do `next unless SiteSetting.try(:assign_enabled)` you can avoid the rescue of `NoMethodError`

eviltrout (approved):

plugin.rb
jjaffeux: I already changed it before your review, but I think you didn't see it

plugin.rb
```diff
@@ -575,4 +575,28 @@ def self.enqueue_assign_hooks(event, payload)
...
+      version 1
+
+      triggerables %i[point_in_time recurring]
+
+      script do |context, fields|
+        next unless SiteSetting.assign_enabled?
+
+        next unless group_id = fields.dig('assignees_group', 'group_id')
+        next unless group = Group.find_by(id: group_id)
+        assign_to = group.group_users.order(Arel.sql('RANDOM()')).first.user
```
martin-brennan, line 593: This is already merged so it doesn't matter much, just curious why you used `Arel.sql('RANDOM()')` instead of just `'RANDOM()'` inside `order()` here @jjaffeux ?

jjaffeux: > This is already merged so it doesn't matter much, just curious why you used `Arel.sql('RANDOM()')` instead of just `'RANDOM()'` inside `order()` here @jjaffeux ?

https://stackoverflow.com/questions/48897070/deprecation-warning-dangerous-query-method-random-record-in-activerecord-5

Is it not the case anymore?

martin-brennan: I didn’t see any warnings when I tried it in rails console @jjaffeux .  Weird.  Anyway no need to change was just curious, thanks for the link!

PR #163, opened 2021-06-27
DEV: Post-release cleanup

CvX:
Included:

* DEV: Post-release cleanup
* DEV: Tests cleanup
* DEV: Import cleanup
* DEV: Drop a compatibility fix
* DEV: Use index_by
* DEV: Use ember imports
* DEV: Use discourseDebounce
* DEV: Use @action
* DEV: Use the optional chaining operator
* DEV: Fix an invalid test
* DEV: Use `discourseModule`
* DEV: Add .prettierrc
* DEV: Sync up the version number


eviltrout (approved):
I love this. It looks like all tests are passing so it's theoretically safe to merge. I would only ask that you are around post merge in case we need to revert anything?

PR #166, opened 2021-06-30
FEATURE: improves random assign automation

jjaffeux:
- Makes a best attempt at being random and assigning people who haven’t been assigned for a long time
- Uses timezones and holidays
- Allows to define a minimum delay between assignments
- Creates a post if no one is available

### DIFF
--- a/lib/random_assign_utils.rb
+++ b/lib/random_assign_utils.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+class RandomAssignUtils
+  def self.user_tzinfo(user_id)
+    timezone = UserOption.where(user_id: user_id).pluck(:timezone).first || "UTC"
+
+    tzinfo = nil
+    begin
+      tzinfo = ActiveSupport::TimeZone.find_tzinfo(timezone)
+    rescue TZInfo::InvalidTimezoneIdentifier
+      Rails.logger.warn("#{User.find_by(id: user_id)&.username} has the timezone #{timezone} set, we do not know how to parse it in Rails (assuming UTC)")
+      timezone = "UTC"
+      tzinfo = ActiveSupport::TimeZone.find_tzinfo(timezone)
+    end
+
+    tzinfo
+  end
+
+  def self.no_one!(topic_id, group)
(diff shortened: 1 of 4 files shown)


jjaffeux:
@SamSaffron It's sadly very hard to test one plugin behavior relying on another plugin. Any idea on this?

lib/random_assign_utils.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+class RandomAssignUtils
+  def self.user_tzinfo(user_id)
```
eviltrout, line 4: I wonder if this method would be useful to have in core? It's a fair amount of logic and seems reusable.

plugin.rb
```diff
@@ -567,14 +570,74 @@ def self.enqueue_assign_hooks(event, payload)
...
-        next unless topic = Topic.find_by(id: topic_id)
+        min_hours = (fields.dig('minimum_time_between_assignments', 'text') || 12).to_i
+        if TopicCustomField
+            .where(name: 'assigned_to_id', topic_id: topic_id)
+            .where('created_at < ?', min_hours.hours.ago)
+            .exists?
+          next
+        end
+
+        users_on_holiday = User
```
eviltrout, line 587: I recommend using a `Set.new` for this. It'll make the `include?` below significantly faster if there are many users.

lib/random_assign_utils.rb
jjaffeux: Sure, at the moment it's in dev-additions and I ported it over here.

Where would you see it defined?

```
User.tzinfo(user_id)
UserOption.tzinfo(user_id)
```

?

eviltrout: I'd say `UserOption` to keep `User` shorter.

SamSaffron:
I think given this design we have I think CI is going to have to always include Discourse Automation, when testing any plugin. 

We should do a TODO to discuss this problem. Maybe we need metadata on plugins that specifies which plugins to load when running tests.

eviltrout (approved):

PR #169, opened 2021-07-11
DEV: Move assignments from custom fields to a table

CvX:


CvX:
@ZogStriP The PR is now in a reviewable state. The only question I have is regarding this feature: https://github.com/discourse/discourse-assign/commit/23aa28e2b4c38995d25bd3b71c95a44469f34f3a That's the only thing that's using topic custom fields.

I kept it as is, since I don't thing adding support for this edge case to the new `Assignment` model path is worth it. Alternatively I could also just remove it. 😜 Should I keep it?

app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -33,18 +32,20 @@ def claim
...
-        "topic_id = :topic_id AND name = 'assigned_to_id' AND value IS NOT NULL",
-        topic_id: topic_id
-      ).pluck(:value)
+      assigned = Assignment
+        .where(topic_id: topic_id)
+        .not(assigned_to_id: nil)
+        .pluck_first(:assigned_to_id)
 
-      if assigned && user_id = assigned[0]
+      if assigned && user_id = assigned
```
ZogStriP, line 40: This could become

```rails
user_id = Assignment...

if user_id
```

app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -33,18 +32,18 @@ def claim
...
-      ).pluck(:value)
+      assigned_id = Assignment
+        .where(topic_id: topic_id)
+        .not(assigned_to_id: nil)
+        .pluck_first(:assigned_to_id)
 
-      if assigned && user_id = assigned[0]
-        extras = nil
-        if user = User.where(id: user_id).first
+      if assigned
```
ZogStriP, line 40: I think you meant `assigned_id` here?

```suggestion
      if assigned_id
```

CvX: What a nice opportunity to write a spec 😆

db/migrate/20210627100932_create_assignments.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class CreateAssignments < ActiveRecord::Migration[6.1]
+  def change
+    create_table :assignments do |t|
+      t.integer :topic_id, null: false
+      t.integer :assigned_to_id, null: false
```
ZogStriP, line 7: How are you planning on using this to differentiate between `users` and `groups`?

CvX: A polymorphic `has_one` association, using a (to-be-added) `assigned_to_type` column.

ZogStriP (approved):

PR #171, opened 2021-07-14
FIX: CF to table migration was incorrect

CvX:
The order of values in the `INSERT` query was invalid (`assigned_to_id` and `topic_id` were swapped)

### DIFF
--- a/db/post_migrate/20210714173022_correctly_move_assignments_from_custom_fields_to_a_table.rb
+++ b/db/post_migrate/20210714173022_correctly_move_assignments_from_custom_fields_to_a_table.rb
@@ -2,6 +2,8 @@
 
 class CorrectlyMoveAssignmentsFromCustomFieldsToATable < ActiveRecord::Migration[6.1]
   def up
+    # An old version of 20210709101534 incorrectly imported `assignments` with
+    # the topic_id and assigned_to_id columns flipped. This query deletes those invalid records.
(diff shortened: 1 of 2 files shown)


db/post_migrate/20210714173022_correctly_move_assignments_from_custom_fields_to_a_table.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class CorrectlyMoveAssignmentsFromCustomFieldsToATable < ActiveRecord::Migration[6.1]
+  def up
+    execute <<~SQL
```
davidtaylorhq, line 5: Maybe add a comment to explain what this query is doing? It would be a bit confusing without the context. Something like

```suggestion
    # An old version of 20210709101534 incorrectly imported `assignments` with
    # the topic_id and assigned_to_id columns flipped. This query deletes those invalid records.
    execute <<~SQL
```

davidtaylorhq (approved):

db/post_migrate/20210714173022_correctly_move_assignments_from_custom_fields_to_a_table.rb
CvX: Oh, good one!

davidtaylorhq (approved):

PR #176, opened 2021-07-22
FIX: filter by "nobody" was broken

techAPJ:
Meta ref: https://meta.discourse.org/t/filter-by-nobody-seems-broken/197770/


CvX (approved):
Thank you for the fix **and** the spec! 😃

PR #178, opened 2021-07-30
DEV: Remove old review queue code

CvX:
I think it should have been deleted with https://github.com/discourse/discourse-assign/pull/47? That's where e.g. `assign_locks_flags` site setting was removed.


romanrizzi (approved):
I totally missed this code. Good catch!

PR #179, opened 2021-07-30
FIX: Hide Unassigned if user doesn't have access

CvX:
There's no need to display the "Unassigned" button for regular users when `assigns_public` is `false`.


AndrewPrigorshnev (approved):
LGTM 👍

CvX:
(updating the names with new prefixes 😉)

PR #180, opened 2021-07-30
DEV: add a prefix with the plugin name to spec names

AndrewPrigorshnev:
I had discourse-assign installed on my local Discourse instance and one spec that has the same name as another spec in Core was causing the error when running qunit tests locally, the whole suite didn't want to run:

Screenshot 2021-07-30 at 20 12 57

I've added a prefix to that spec (https://github.com/discourse/discourse-assign/commit/3aeeea3c8af13e5ee9c0fee6f5ae7b2c5c371241) and to the rest of them to avoid conflicts and make it easier to filter specs for the plugin during local development.


CvX (approved):
Ha, I changed that also here: https://github.com/discourse/discourse-assign/pull/179/files#diff-82570edaa4996d0753afb172b1a73445c225fa12c97d65ff9812995de4f16618R10 😉 

But the general cleanup of names is good to see!

PR #192, opened 2021-08-29
FIX: unassign/assign when group pm is archived

lis2:
Send correct messages and correctly unassign/assign to user/group if group PM is archived or moved back to inbox.

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -566,8 +566,15 @@ after_initialize do
     assigned_to_id = topic.assignment&.assigned_to_id
     assigned_to_type = topic.assignment&.assigned_to_type
 
-    if info[:user]&.id == assigned_to_id && assigned_to_type == "User"
-      TopicTrackingState.publish_assigned_private_message(topic, assigned_to_id)
+    assigned_to_ids =
+      if info[:user]&.id == assigned_to_id && assigned_to_type == "User"
+        [assigned_to_id]
+      elsif info[:group]&.id == assigned_to_id && assigned_to_type == "Group"
+        Group.find(assigned_to_id).users.pluck(:id)
+      end
+
+    assigned_to_ids && assigned_to_ids.each do |user_id|
+      TopicTrackingState.publish_assigned_private_message(topic, user_id)
     end
 
     next if !SiteSetting.unassign_on_group_archive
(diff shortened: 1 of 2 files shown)


lis2:
Hey @davidtaylorhq - than you for your quick patch on Friday. I fixed that logic a bit and added specs, can I ask you to review?

plugin.rb
```diff
@@ -521,8 +521,15 @@ class ::ListController
     assigned_to_id = topic.assignment&.assigned_to_id
     assigned_to_type = topic.assignment&.assigned_to_type
 
-    if info[:user]&.id == assigned_to_id && assigned_to_type == "User"
-      TopicTrackingState.publish_assigned_private_message(topic, assigned_to_id)
+    assigned_to_ids =
+      if info[:user]&.id == assigned_to_id && assigned_to_type == "User"
+        [assigned_to_id]
+      elsif info[:group]&.id == assigned_to_id && assigned_to_type == "Group"
+        Group.find(assigned_to_id).users.pluck(:id)
```
martin-brennan, line 528: Should we maybe have a check to make sure the group is not deleted? Maybe it's not necessary if `info[:group]` is an actual Group record

plugin.rb
```diff
@@ -521,8 +521,15 @@ class ::ListController
...
+      end
+
+    assigned_to_ids && assigned_to_ids.each do |user_id|
```
martin-brennan, line 531: Could just be `assigned_to_ids&.each`

martin-brennan (approved):
Just some minor comments otherwise looks good

plugin.rb
tgxworld: This is going to take two queries to execute. Can we get the user ids from the `GroupUser` table instead?

plugin.rb
```diff
@@ -548,21 +555,26 @@ class ::ListController
     assigned_to_id = topic.assignment.assigned_to_id
     assigned_to_type = topic.assignment.assigned_to_type
 
-    if info[:user]&.id == assigned_to_id && assigned_to_type == "User"
-      TopicTrackingState.publish_assigned_private_message(topic, assigned_to_id)
+    assigned_to_ids =
+      if info[:user]&.id == assigned_to_id && assigned_to_type == "User"
+        [assigned_to_id]
+      elsif info[:group]&.id == assigned_to_id && assigned_to_type == "Group"
+        Group.find(assigned_to_id).users.pluck(:id)
```
tgxworld, line 562: Ditto https://github.com/discourse/discourse-assign/pull/192/files#r704071913

spec/integration/assign_spec.rb
```diff
@@ -54,6 +57,47 @@ def assert_publish_topic_state(topic, user)
...
+    end
+
+    it "unassign and assign user if unassign_on_group_archive" do
+      SiteSetting.unassign_on_group_archive = true
+      assigner = TopicAssigner.new(pm, user)
+      assigner.assign(user)
+
+      GroupArchivedMessage.archive!(group.id, pm.reload)
+      expect(pm.assignment).to eq(nil)
+      expect(pm.custom_fields["prev_assigned_to_id"]).to eq(user.id.to_s)
```
tgxworld, line 81: Hmm did we not register this as an integer custom field? Odd that we need to do `to_s`

plugin.rb
tgxworld: I think we need to redo publishing for groups. If a group has 1000 users, this is going to send 1000 messages. Instead, it will be better if we can just publish the a group channel.

plugin.rb
```diff
@@ -490,14 +491,21 @@ class ::ListController
...
+        { topic_id: topic.id },
+        user_ids: [assignee.id]
+      )
+    else
+      MessageBus.publish(
+        "/private-messages/assigned",
+        { topic_id: topic.id },
+        group_ids: [assignee.id]
+      )
+    end
```
tgxworld, line 508:
```suggestion
    opts = 
      if assignee.is_a?(User)
        { user_ids: [assignee.id] }
      else
        { group_ids: [assignee.id] }
      end

    MessageBus.publish(
      "/private-messages/assigned",
      { topic_id: topic.id },
      opts
    )
```

spec/integration/assign_spec.rb
```diff
@@ -23,37 +23,82 @@
...
+      assigner = TopicAssigner.new(pm, user)
+      assigner.assign(user)
+
+      GroupArchivedMessage.archive!(group.id, pm.reload)
+      expect(pm.assignment).to eq(nil)
+      expect(pm.custom_fields["prev_assigned_to_id"]).to eq(user.id)
+      expect(pm.custom_fields["prev_assigned_to_type"]).to eq("User")
+
+      GroupArchivedMessage.move_to_inbox!(group.id, pm.reload)
+      expect(pm.assignment.assigned_to).to eq(user)
```
tgxworld, line 86: Should we assert that the custom fields have been cleared as well?

tgxworld (approved):

PR #195, opened 2021-09-02
FEATURE: new assignable group option instead of messageable

lis2:
V1 of group assign was using "who can message" to determine if group can be assigned.

This PR is introducing new separate setting "who can assign"

Correlated PR https://github.com/discourse/discourse/pull/14216 (Specs will fail until core pr is merged)

Screen Shot 2021-09-02 at 11 06 03 am

### DIFF
--- a/db/migrate/20210830024453_add_assignable_level_to_groups.rb
+++ b/db/migrate/20210830024453_add_assignable_level_to_groups.rb
@@ -2,6 +2,6 @@
 
 class AddAssignableLevelToGroups < ActiveRecord::Migration[6.1]
   def change
-    add_column :groups, :assignable_level, :integer, default: 0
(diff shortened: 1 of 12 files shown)


db/migrate/20210830024453_add_assignable_level_to_groups.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddAssignableLevelToGroups < ActiveRecord::Migration[6.1]
+  def change
+    add_column :groups, :assignable_level, :integer, default: 0
```
eviltrout, line 5: Since there is a default should this be `null: false`?

eviltrout (approved):

PR #197, opened 2021-09-03
FEATURE: better UI for group assignments

lis2:
On the group assignment page, we should be able to display
- all topics assigned to that group + to user belonging to that group
- all topics assigned directly to group
- all topics assigned to individual user

https://user-images.githubusercontent.com/72780/131953304-aae4ad9d-e4fa-4ae8-9af2-47be9ac2eabe.mov

### DIFF
--- a/app/controllers/discourse_assign/assign_controller.rb
+++ b/app/controllers/discourse_assign/assign_controller.rb
@@ -127,7 +127,7 @@ module DiscourseAssign
       end
 
       group_assignment_count = Topic
-        .joins("JOIN assignments a ON a.topic_id = topics.id AND a.assigned_to_id IS NOT NULL ")
+        .joins("JOIN assignments a ON a.topic_id = topics.id")
         .where(<<~SQL, group_id: group.id)
           a.assigned_to_id = :group_id AND a.assigned_to_type = 'Group'
         SQL
@@ -135,7 +135,7 @@ module DiscourseAssign
         .count
 
       assignment_count = Topic
-        .joins("JOIN assignments a ON a.topic_id = topics.id AND a.assigned_to_id IS NOT NULL ")
+        .joins("JOIN assignments a ON a.topic_id = topics.id")
         .where(<<~SQL, group_id: group.id)
           a.assigned_to_id IN (SELECT group_users.user_id FROM group_users WHERE group_id = :group_id) AND a.assigned_to_type = 'User'
         SQL
--- a/app/controllers/discourse_assign/assign_controller.rb
+++ b/app/controllers/discourse_assign/assign_controller.rb
@@ -131,15 +131,13 @@ module DiscourseAssign
         .where(<<~SQL, group_id: group.id)
           a.assigned_to_id = :group_id AND a.assigned_to_type = 'Group'
         SQL
-        .where("topics.deleted_at IS NULL")
         .count
 
       assignment_count = Topic
         .joins("JOIN assignments a ON a.topic_id = topics.id")
-        .where(<<~SQL, group_id: group.id)
-          a.assigned_to_id IN (SELECT group_users.user_id FROM group_users WHERE group_id = :group_id) AND a.assigned_to_type = 'User'
-        SQL
-        .where("topics.deleted_at IS NULL")
+        .joins("JOIN group_users ON group_users.user_id = a.assigned_to_id ")
+        .where("group_users.group_id = ?", group.id)
+        .where("a.assigned_to_type = 'User'")
         .count
 
       render json: {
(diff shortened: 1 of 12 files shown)


app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -126,17 +126,27 @@ def group_members
         SQL
       end
 
+      group_assignment_count = Topic
+        .joins("JOIN assignments a ON a.topic_id = topics.id AND a.assigned_to_id IS NOT NULL ")
```
eviltrout, line 130: `a.assigned_to_id IS NOT NULL` appears redundant because of the `where a.assigned_to_id = :group_id` below

assets/javascripts/discourse/templates/components/group-assigned-filter.hbs
```diff
@@ -13,6 +13,19 @@
       {{filter.assignments_count}}
     </div>
   {{/link-to}}
+{{else if group-name}}
```
eviltrout, line 16: We prefer `camelCase` for variables in Javascript/HBS. The exception is component names which should have a dash.

app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -126,17 +126,27 @@ def group_members
         SQL
       end
 
+      group_assignment_count = Topic
+        .joins("JOIN assignments a ON a.topic_id = topics.id")
+        .where(<<~SQL, group_id: group.id)
+          a.assigned_to_id = :group_id AND a.assigned_to_type = 'Group'
+        SQL
+        .where("topics.deleted_at IS NULL")
```
tgxworld, line 134: I don't think we need this since there is a default scope on the topic model.

app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -126,17 +126,27 @@ def group_members
         SQL
       end
 
+      group_assignment_count = Topic
+        .joins("JOIN assignments a ON a.topic_id = topics.id")
+        .where(<<~SQL, group_id: group.id)
+          a.assigned_to_id = :group_id AND a.assigned_to_type = 'Group'
```
tgxworld, line 132: Do we have an index on the columns here? Otherwise, filtering through all the topics is going to be expensive on a large forum.

app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -126,17 +126,27 @@ def group_members
...
+        .count
+
       assignment_count = Topic
-        .joins("JOIN assignments a ON a.topic_id = topics.id AND a.assigned_to_id IS NOT NULL ")
+        .joins("JOIN assignments a ON a.topic_id = topics.id")
         .where(<<~SQL, group_id: group.id)
-          (a.assigned_to_id IN (SELECT group_users.user_id FROM group_users WHERE group_id = :group_id) AND a.assigned_to_type = 'User')
-          OR
-          (a.assigned_to_id = :group_id AND a.assigned_to_type = 'Group')
+          a.assigned_to_id IN (SELECT group_users.user_id FROM group_users WHERE group_id = :group_id) AND a.assigned_to_type = 'User'
```
tgxworld, line 140: Now that we don't have an `OR`, I think we can just do a LEFT JOIN on `group_users` which should be way more efficient.

assets/javascripts/discourse-assign/routes/group-assigned.js.es6
```diff
@@ -14,6 +14,10 @@ export default DiscourseRoute.extend({
       group: this.modelFor("group"),
     });
     controller.group.set("assignment_count", model.assignment_count);
+    controller.group.set(
```
tgxworld, line 17: Let's combine this line with the line above by using `setProperties` to update multiple attributes.

spec/components/topic_query_spec.rb
```diff
@@ -19,56 +19,73 @@
...
 
-      expect(assigned_messages).to contain_exactly(@private_message, @topic)
+      expect(assigned_messages).to contain_exactly(private_message, topic, group_topic)
     end
 
     it 'Excludes topics and PMs not assigned to user' do
       assigned_messages = TopicQuery.new(user2, { page: 0 }).list_messages_assigned(user2).topics
 
-      expect(assigned_messages).to be_empty
+      expect(assigned_messages).to eq([group_topic])
```
tgxworld, line 41:
```suggestion
      expect(assigned_messages).to contain_exactly(group_topic)
```

tgxworld (approved):

PR #198, opened 2021-09-08
FEATURE: advance search groups

lis2:
Allow advance search filter to select topics assigned to specific group

Screen Shot 2021-09-08 at 11 29 46 am


lis2:
@tgxworld I think you will like how easy it was to do with our last modifications :)

tgxworld (approved):

PR #199, opened 2021-09-08
FIX: add assigned fields to suggested topic serializer

lis2:
With hard refresh, information about assigned user disappeared from suggested topics.

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -201,16 +201,16 @@ after_initialize do
         assignments = Assignment.strict_loading.where(topic: topics)
         assignments_map = assignments.index_by(&:topic_id)
 
-        user_ids = assignments.filter { |assignment| assignment.assigned_to_type == "User" }.map(&:assigned_to_id)
+        user_ids = assignments.filter { |assignment| assignment.assigned_to_user? }.map(&:assigned_to_id)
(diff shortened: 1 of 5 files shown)


tgxworld:
Can we add a test here? :)

lis2:
@tgxworld done :)

plugin.rb
```diff
@@ -405,6 +405,23 @@ class ::ListController
     (SiteSetting.assigns_public || scope.can_assign?) && object.topic.assigned_to&.is_a?(Group)
   end
 
+  # SuggestedTopic serializer
+  add_to_serializer(:suggested_topic, :assigned_to_user, false) do
+    DiscourseAssign::Helpers.build_assigned_to_user(object.assigned_to, object)
+  end
+
+  add_to_serializer(:suggested_topic, :include_assigned_to_user?) do
+    (SiteSetting.assigns_public || scope.can_assign?) && object.assigned_to&.is_a?(User)
```
tgxworld, line 414: Can `object.assigned_to_id` be `nil`? If so, it'll be more efficient to check for the present of that before fetching the `ActiveRecord` object. Also, can't we determine whether it is a user or not from the `assign_type` instead?

tgxworld (approved):

PR #204, opened 2021-09-14
FIX: Allow Never selection for frequency of assigned topic reminders

martin-brennan:
If the user selected Never for their frequency of assigned topic
reminder PMs, the option automatically set back to the default
value (Monthly) because of the user of the `or` computed helper,
which considered the 0 value of Never to be falsey.

This is the same problem that occurred in core which was fixed by
https://github.com/discourse/discourse/commit/d3779d4cf75bd9c21cf61e016047a92fe66d3fdd

We need to be careful around using the `or` helper if any of the
options could be considered falsey.


romanrizzi (approved):
Awesome! 👍

PR #212, opened 2021-09-16
FIX: Refresh first post after assign and unassign

nbianca:
The first post showed the assign status when it was loaded and did not
update if it changed. The page had to be refreshed for the text to be
updated.


ZogStriP (approved):

SamSaffron:
nice fix!

PR #216, opened 2021-09-28
FIX: n+1 query on list of private messages assigned

lis2:
When getting a list of private messages assigned to groups/users, we should include :allowed_users to avoid n+1

Screen Shot 2021-09-29 at 8 54 47 am


tgxworld (approved):
LGTM :+1:

PR #227, opened 2021-11-02
FEATURE: display indirect assignments in the first post

lis2:
A link in the first post will allow going to a specific post assigned to a user or group.

Screen Shot 2021-11-02 at 2 55 35 pm

### DIFF
--- a/assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
+++ b/assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
@@ -13,7 +13,7 @@ import TopicButtonAction, {
 } from "discourse/controllers/topic-bulk-actions";
 import { inject } from "@ember/controller";
 import I18n from "I18n";
-import { get } from "@ember/object";
+import { isEmpty } from "@ember/utils";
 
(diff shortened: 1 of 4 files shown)


assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
```diff
@@ -336,6 +336,77 @@ function initialize(api) {
     },
   });
 
+  api.createWidget("assigned-to-first-post", {
+    html(attrs) {
+      const topic = attrs.topic;
+      const [assignedToUser, assignedToGroup, indirectlyAssignedTo] = [
```
eviltrout, line 342: Small tip: You could destructure this like `{ assignedToUser, assignedToGroup, indirectlyAssignedTo } = topic`;

eviltrout (approved):

assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
lis2: Actually, in this case it will not work as serializer is returning `assigned_to_user`, and we want to have JS style variable `assignedToUser`

PR #229, opened 2021-11-02
FIX: move post assignment when move post

lis2:
When an assigned post is moved to the new topic, it becomes topic assignment.

When an assigned post is moved to an existing topic it stays post assignment

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -737,6 +737,16 @@ after_initialize do
     end
   end
 
+  on(:post_moved) do |post, original_topic_id|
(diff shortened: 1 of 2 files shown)


plugin.rb
```diff
@@ -737,6 +737,16 @@ class ::ListController
     end
   end
 
+  on(:post_moved) do |post, original_topic_id|
+    assignment = Assignment.where(topic_id: original_topic_id, target_type: "Post", target_id: post.id).first
+    return if !assignment
```
eviltrout, line 742:
```suggestion
    next if !assignment
```

Return in a block is probably not what you want here, it returns from the higher scope.

eviltrout (approved):

PR #232, opened 2021-11-04
FEATURE: small post action with link to post

lis2:
When a post is assigned/unassigned, we want small action to have a link to the specific post.

Screen Shot 2021-11-04 at 2 32 05 pm

Related PR: https://github.com/discourse/discourse/pull/14816

### DIFF
--- a/lib/assigner.rb
+++ b/lib/assigner.rb
@@ -250,13 +250,19 @@ class ::Assigner
       end
     end
     if !silent
+      custom_fields = { "action_code_who" => assign_to.is_a?(User) ? assign_to.username : assign_to.name }
+
+      if post_target?
+        custom_fields.merge!("action_code_href" => "#{topic.url}/#{@target.post_number}")
+      end
+
       topic.add_moderator_post(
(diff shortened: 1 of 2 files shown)


lib/assigner.rb
```diff
@@ -250,13 +250,19 @@ def assign(assign_to, silent: false)
       end
     end
     if !silent
+      custom_fields = { "action_code_who" => assign_to.is_a?(User) ? assign_to.username : assign_to.name }
+
+      if post_target?
+        custom_fields.merge!("action_code_href" => "/p/#{@target.id}")
```
eviltrout, line 256: I don't think this will work with subfolder installs since we aren't using `getURL`. Should the core PR be updated to support them?

lis2: Great catch! thank you, I adjusted core PR:
`https://github.com/discourse/discourse/pull/14816/commits/b2f5bd2c6863283b25371815f3e1b18bdddae1fb`

eviltrout (approved):

PR #249, opened 2021-11-17
Show custom topic tracking message when the user 'never' automatically tracks topics and gets assigned

nattsw:
When the user "never" tracks topics, 
Screenshot 2021-11-17 at 3 08 59 PM

We want to show a more accurate tracking message when he gets assigned to a topic
Screenshot 2021-11-17 at 3 10 09 PM

### DIFF
--- a/assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
+++ b/assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
@@ -511,25 +511,13 @@ function initialize(api) {
       "topic",
       "topic.details.{notification_level,notifications_reason_id}"
     )
-    notificationReasonText(topic, topicDetails) {
-      if (this.currentUser.auto_track_topics_after_msecs === -1) {
-        if (
-          topic.assigned_to_user &&
-          topic.assigned_to_user.username === this.currentUser.username
-        ) {
-          return I18n.t("notification_reason.user");
-        }
-        if (
-          topic.assigned_to_group &&
-          this.currentUser.groups &&
-          this.currentUser.groups.some(
-            ({ id }) => id === topic.assigned_to_group.id
-          )
-        ) {
-          return I18n.t("notification_reason.group", {
-            name: topic.assigned_to_group.name,
-          });
-        }
+    notificationReasonText(topic) {
+      if (
+        this.currentUser.auto_track_topics_after_msecs === -1 &&
+        topic.assigned_to_user &&
+        topic.assigned_to_user.username === this.currentUser.username
+      ) {
+        return I18n.t("notification_reason.user");
       }
 
       return this._super(...arguments);
--- a/assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
+++ b/assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
@@ -507,6 +507,8 @@ function initialize(api) {
   });
 
   api.modifyClass("component:topic-notifications-button", {
+    pluginId: PLUGIN_ID,
+
     @discourseComputed(
       "topic",
       "topic.details.{notification_level,notifications_reason_id}"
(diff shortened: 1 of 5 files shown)


assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
```diff
@@ -506,6 +506,24 @@ function initialize(api) {
...
   });
 
+  api.modifyClass("component:topic-notifications-button", {
+    @discourseComputed(
+      "topic",
+      "topic.details.{notification_level,notifications_reason_id}"
+    )
+    notificationReasonText(topic) {
+      if (
+        this.currentUser.auto_track_topics_after_msecs === -1 &&
```
tgxworld, line 516: The `-1` is sort of like a magic number. I'm wondering if we can move this check to the server side where we ship down a boolean field `auto_track_disabled` instead. 

```
  add_to_serializer(:current_user, :auto_track_disabled) do
    user.user_option.auto_track_topics_after_msecs == AutoTrackDurationSiteSetting.values[:never]
  end
```

^^ The above is a sample of what I'm thinking about and the code might not work.

assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
```diff
@@ -506,6 +506,24 @@ function initialize(api) {
     },
   });
 
+  api.modifyClass("component:topic-notifications-button", {
+    @discourseComputed(
+      "topic",
+      "topic.details.{notification_level,notifications_reason_id}"
+    )
+    notificationReasonText(topic) {
```
tgxworld, line 514: Usually we try not override functions since any changes in core might inadvertently break this override. What we usually do is to extend core by providing an API for us to be able to register a callback in order to achieve the results we want.

test/javascripts/acceptance/never-track-assign-reason-test.js.es6
```diff
@@ -0,0 +1,70 @@
...
+  function (needs) {
+    needs.user();
+    needs.settings({
+      assign_enabled: true,
+      assigns_user_url_path: "/",
+    });
+
+    needs.pretender((server, helper) => {
+      server.get("/t/44.json", () => {
+        let topic = cloneJSON(topicFixtures["/t/130.json"]);
```
tgxworld, line 22: O nice :+1: I didn't know we have a `cloneJSON` helper function before this.

assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
nattsw: I could add a `auto_track_disabled` into the serializer instead of this setting.

assets/javascripts/discourse-assign/initializers/extend-for-assigns.js.es6
nattsw: I'm not sure how I feel about this.

> override functions since any changes in core

I would think the API should not include `modifyClass` if the intention is for API users not to override functions.

> extend core by providing an API

wouldn't it be a bit of an overkill to have to create a new API each time we want to make a modification from a plugin? I understand overriding it this way would result in some coupling, but extending core each time we should not override a function feels like not a good long term solution. 🤔

nattsw: Another thought would be that plugins should have some form of compatibility for versions of Discourse. Having version control feels like a better long term solution, where plugin authors can freely use available APIs and be sure that it works with their current version of the app.

Them having a dependency on the core API change also doesn't seem right. It's a subjective problem.

tgxworld: We discussed this privately and we feel like adding an extension to core for this small feature isn't worth it as this point. Since it isn't the end of the world if the description goes missing if the function override stops working, we're OK with this overriding the function here.

test/javascripts/acceptance/never-track-assign-reason-test.js.es6
martin-brennan: Yes very handy for modifying fixtures, a helper for this combined with some sort of auto generation of fixture data + a contract would make fixtures way less work for us.

martin-brennan (approved):
LGTM :+1:

PR #250, opened 2021-11-17
FIX: Refactor assign / reassign to combine logic

janzenisaac:
Fixes: https://dev.discourse.org/t/self-assign-is-not-working-on-mobile/56331/21?u=isaacjanzen

- Combines assign and reassign logic... removing bulky and extra logic

### DIFF
--- a/app/controllers/discourse_assign/assign_controller.rb
+++ b/app/controllers/discourse_assign/assign_controller.rb
@@ -56,8 +56,6 @@ module DiscourseAssign
       target = target_type.constantize.where(id: target_id).first
       raise Discourse::NotFound unless target
 
-      # perhaps?
-      #Scheduler::Defer.later "assign topic" do
       assign = Assigner.new(target, current_user).assign(assign_to)
 
       if assign[:success]
--- a/app/controllers/discourse_assign/assign_controller.rb
+++ b/app/controllers/discourse_assign/assign_controller.rb
@@ -44,11 +44,27 @@ module DiscourseAssign
     end
 
     def assign
-      reassign_or_assign_target(action: "assign")
-    end
+      target_id = params.require(:target_id)
+      target_type = params.require(:target_type)
+      username = params.permit(:username)['username']
+      group_name = params.permit(:group_name)['group_name']
 
-    def reassign
-      reassign_or_assign_target(action: "reassign")
+      assign_to = username.present? ? User.find_by(username_lower: username.downcase) : Group.where("LOWER(name) = ?", group_name.downcase).first
+
+      raise Discourse::NotFound unless assign_to
+      raise Discourse::NotFound if !Assignment.valid_type?(target_type)
+      target = target_type.constantize.where(id: target_id).first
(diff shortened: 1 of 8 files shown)


janzenisaac:
This feels like a flaky test... @jjaffeux What do you think?

jjaffeux:
> This feels like a flaky test... @jjaffeux What do you think?

I think this is an issue given it's happening all the time :po

app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -44,11 +44,27 @@ def unassign
...
+      assign_to = username.present? ? User.find_by(username_lower: username.downcase) : Group.where("LOWER(name) = ?", group_name.downcase).first
+
+      raise Discourse::NotFound unless assign_to
+      raise Discourse::NotFound if !Assignment.valid_type?(target_type)
+      target = target_type.constantize.where(id: target_id).first
+      raise Discourse::NotFound unless target
+
+      # perhaps?
+      #Scheduler::Defer.later "assign topic" do
+      assign = Assigner.new(target, current_user).send("assign", assign_to)
```
ZogStriP, line 61: Why are you using `send` when you can call the `assign` method directly?

```suggestion
      assign = Assigner.new(target, current_user).assign(assign_to)
```

app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -44,11 +44,27 @@ def unassign
...
-      reassign_or_assign_target(action: "reassign")
+      assign_to = username.present? ? User.find_by(username_lower: username.downcase) : Group.where("LOWER(name) = ?", group_name.downcase).first
+
+      raise Discourse::NotFound unless assign_to
+      raise Discourse::NotFound if !Assignment.valid_type?(target_type)
+      target = target_type.constantize.where(id: target_id).first
+      raise Discourse::NotFound unless target
+
+      # perhaps?
+      #Scheduler::Defer.later "assign topic" do
```
ZogStriP, line 60: Is the `assign` method a CPU/DB intensive process?

app/controllers/discourse_assign/assign_controller.rb
janzenisaac: This was left over from an old method. Good catch, I will remove.

app/controllers/discourse_assign/assign_controller.rb
janzenisaac: From looking at the logs on assign, it does not seem overly heavy... this was a comment (left over) from before I started working on the re-assign workflow, so i figured I would leave it in. In the case someone was coming back to update? 🤷‍♂️

ZogStriP: I'd rather we remove the comment to keep the code clean (it'll be in the git history anyways). We can always revisit this later if we end up hitting a performance wall.

janzenisaac:
@ZogStriP could i get a re-review on this?

eviltrout (approved):

PR #261, opened 2021-12-04
DEV: use the empty-state component

AndrewPrigorshnev:


CvX (commented):
I don't think this component is in a stable release yet. You'll need to add an entry to .discourse-compatibility

CvX (approved):

PR #263, opened 2021-12-07
FEATURE: active flag for assignments

lis2:
When topic is closed, we mark assignments as "inactive". Then when it is reopen and setting reassign_on_open is enabled, we bring back previous assignments.

We were already using custom fields for archive message and move to inbox. I changed custom fields solution to use active flag on Assignment model

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -85,7 +85,7 @@ after_initialize do
         ON topics.id = a.topic_id AND a.assigned_to_id IS NOT NULL
       SQL
       .where(<<~SQL, group_id: object.id)
-        a.active IS TRUE AND
+        a.active AND
         ((
           a.assigned_to_type = 'User' AND a.assigned_to_id IN (
             SELECT group_users.user_id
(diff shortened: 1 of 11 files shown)


plugin.rb
```diff
@@ -284,13 +284,13 @@ class ::Group
 
     if name == "nobody"
       next results
-        .joins("LEFT JOIN assignments a ON a.topic_id = topics.id")
+        .joins("LEFT JOIN assignments a ON a.topic_id = topics.id AND active IS TRUE")
```
eviltrout, line 287: Trivial but in SQL you can do `AND active` without the `IS TRUE`. I would never not approve for this, and it's up to you to use it or not.

plugin.rb
```diff
@@ -619,15 +621,15 @@ class ::ListController
...
+    (SiteSetting.assigns_public || scope.can_assign?) && object.assignment&.assigned_to&.is_a?(User) && object.assignment.active
   end
 
   add_to_serializer(:post, :assigned_to_group, false) do
     object.assignment&.assigned_to
   end
 
   add_to_serializer(:post, 'include_assigned_to_group?') do
-    (SiteSetting.assigns_public || scope.can_assign?) && object.assignment&.assigned_to&.is_a?(Group)
+    (SiteSetting.assigns_public || scope.can_assign?) && object.assignment&.assigned_to&.is_a?(Group) && object.assignmen.active
```
eviltrout, line 632: There's a typo here. The fact it didn't fail the tests tells me we could use more coverage here.

PR #266, opened 2021-12-13
FIX: indirect assigned to in search

lis2:
Search similarly to topic list should return a hash with post_number


davidtaylorhq (approved):

ZogStriP:
No spec has been added/changed? 🤔

PR #272, opened 2021-12-23
🐛 FIX: remove assigned to username property from the payload

Saif-Shines:
Context: https://meta.discourse.org/t/discourse-assign/58044/163?u=saif-shines


SamSaffron:
thanks for the PR, merged  @communiteq's version cause it covered one more line we needed.

PR #273, opened 2021-12-23
Fix: set these fields in condition below only

communiteq:
https://meta.discourse.org/t/discourse-assign/58044/164?u=rgj


SamSaffron (approved):

SamSaffron:
looks good, thanks!

PR #276, opened 2022-01-04
FIX: bulk assign modal was broken

techAPJ:
When trying to bulk assign topics to a user via bulk topics action modal
the modal was broken because the key "model.description" was missing.

This commit adds key "model.description" and adds a new locale for it to
indicate that this action will be performed on multiple topics.

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -56,6 +56,8 @@ en:
       assign_post_modal:
         title: "Assign Post"
         description: "Enter the name of the user you'd like to assign this post"
+      assign_bulk_modal:
+        description: "Enter the name of the user you'd like to assign these topics"
       claim:
         title: "claim"
         help: "Assign topic to yourself"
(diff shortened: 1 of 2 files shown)


nattsw (approved):

config/locales/client.en.yml
```diff
@@ -56,6 +56,8 @@ en:
       assign_post_modal:
         title: "Assign Post"
         description: "Enter the name of the user you'd like to assign this post"
+      assign_bulk_modal:
+        description: "Enter the name of the user you'd like to assign these topics"
```
nattsw, line 60: Looks good but just want to confirm that it should be "topics" not "posts"?

techAPJ: Yes the "topics" is correct here since the bulk assign action can only be performed on topics via bulk topic action modal.

Thanks for the review! :)

nattsw: Yup okay just saw the topic.

PR #281, opened 2022-01-12
FIX: Only active assigns should be included in reminders

jbrw:


lib/pending_assigns_reminder.rb
```diff
@@ -63,7 +63,7 @@ def assigned_topics(user, order:)
     Topic
       .joins(:assignment)
       .select(:slug, :id, :title, :fancy_title, 'assignments.created_at AS assigned_at')
-      .where("assignments.assigned_to_id = ? AND assignments.assigned_to_type = 'User'", user.id)
+      .where("assignments.assigned_to_id = ? AND assignments.assigned_to_type = 'User' AND assignments.active = ?", user.id, true)
```
eviltrout, line 66:
```suggestion
      .where("assignments.assigned_to_id = ? AND assignments.assigned_to_type = 'User' AND assignments.active", user.id)
```
It's slightly nicer to not parameterize `true`

eviltrout (approved):

PR #286, opened 2022-01-27
FIX: Count and show only active assignments

nbianca:
After a topic is closed, the assignment is not deleted but becomes
inactive. When this happened, it was not removed from the assignments
count or list.

### DIFF
--- a/spec/requests/list_controller_spec.rb
+++ b/spec/requests/list_controller_spec.rb
@@ -71,7 +71,7 @@ describe ListController do
       Assignment.where(assigned_to: user, target: topic1).update_all(active: false)
 
       get "/topics/group-topics-assigned/#{get_assigned_allowed_group_name}.json"
-      expect(JSON.parse(response.body)['topic_list']['topics'].map { |t| t['id'] }).to eq([])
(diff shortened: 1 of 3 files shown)


spec/requests/list_controller_spec.rb
```diff
@@ -67,6 +67,13 @@
       expect(JSON.parse(response.body)['topic_list']['topics'].map { |t| t['assigned_to_user']['id'] }).to match_array([user.id])
     end
 
+    it 'returns user-assigned-topics-list of users in the assigned_allowed_group and doesnt include inactive topics' do
+      Assignment.where(assigned_to: user, target: topic1).update_all(active: false)
+
+      get "/topics/group-topics-assigned/#{get_assigned_allowed_group_name}.json"
+      expect(JSON.parse(response.body)['topic_list']['topics'].map { |t| t['id'] }).to eq([])
```
CvX, line 74:
```suggestion
      expect(response.parsed_body['topic_list']['topics']).to be_empty
```

CvX (approved):

ZogStriP (approved):

PR #287, opened 2022-01-28
FIX: better group assign message

lis2:
When topic is assigned to group, then notification should be different to distinguish from direct assignments.

### DIFF
--- a/spec/jobs/regular/assign_notification_spec.rb
+++ b/spec/jobs/regular/assign_notification_spec.rb
@@ -6,7 +6,7 @@ RSpec.describe Jobs::AssignNotification do
   describe '#execute' do
     fab!(:user1) { Fabricate(:user, last_seen_at: 1.day.ago) }
     fab!(:user2) { Fabricate(:user, last_seen_at: 1.day.ago) }
-    fab!(:topic) { Fabricate(:topic) }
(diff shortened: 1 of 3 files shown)


spec/jobs/regular/assign_notification_spec.rb
```diff
@@ -79,13 +79,13 @@ def assert_publish_topic_state(topic, user)
...
         expect(messages.length).to eq(1)
-        expect(messages.first.data[:excerpt]).to eq("assigned you the topic '#{topic.title}'")
+        expect(messages.first.data[:excerpt]).to eq("assigned to #{group.name} the topic '#{topic.title}'")
 
         messages = MessageBus.track_publish("/notification-alert/#{user3.id}") do
           described_class.new.execute({ topic_id: topic.id, post_id: post.id, assigned_to_id: group.id, assigned_to_type: 'Group', assigned_by_id: user1.id, silent: false })
         end
         expect(messages.length).to eq(1)
-        expect(messages.first.data[:excerpt]).to eq("assigned you the topic '#{topic.title}'")
+        expect(messages.first.data[:excerpt]).to eq("assigned to #{group.name} the topic '#{topic.title}'")
```
eviltrout, line 88: It might be better for the test to look up the text rather than hard code it.

eviltrout (approved):

PR #300, opened 2022-03-14
FEATURE: improves random assign to assign to a post

jjaffeux:


lib/random_assign_utils.rb
```diff
@@ -1,6 +1,111 @@
 # frozen_string_literal: true
 
 class RandomAssignUtils
+  def self.log_error(automation, message)
+    raise("[discourse-automation id=#{automation.id}] #{message}.")
+  end
```
ZogStriP, line 6: Is it logging or raising? 🤔 

Should either be `log_error` and do `Rails.logger.error()` or `raise_error` and do `raise`

lib/random_assign_utils.rb
```diff
@@ -1,6 +1,111 @@
...
+    unless group = Group.find_by(id: group_id)
+      log_error(automation, "Group(#{group_id}) not found")
+    end
+
+    users_on_holiday = Set.new(
+      User
+        .where(id:
+          UserCustomField
+          .where(name: 'on_holiday', value: 't')
+          .pluck(:user_id)
```
ZogStriP, line 47: I _think_ the `select` here is better as it will only do one SQL query rather than 2?

```suggestion
          .select(:user_id)
```

lib/random_assign_utils.rb
jjaffeux: will do raie error because I want to interrupt script

lib/random_assign_utils.rb
```diff
@@ -1,6 +1,111 @@
...
+      .reject { |user_id| users_on_holiday.include?(user_id) }
+
+    if group_users_ids.empty?
+      RandomAssignUtils.no_one!(topic_id, group.name)
+      return
+    end
+
+    last_assignees_ids = RandomAssignUtils.recently_assigned_users_ids(
+      topic_id,
+      (fields.dig('max_recently_assigned_days', 'value').presence || 180).to_i.days.ago
```
ZogStriP, line 64: Might be too much on the same line? Maybe extract this line into a variable?

lib/random_assign_utils.rb
```diff
@@ -1,6 +1,111 @@
...
+    )
+    users_ids = group_users_ids - last_assignees_ids
+    if users_ids.blank?
+      recently_assigned_users_ids = RandomAssignUtils.recently_assigned_users_ids(topic_id, (fields.dig('min_recently_assigned_days', 'value').presence || 14).to_i.days.ago)
```
ZogStriP, line 68: Ditto ^^ I would extract the `fields.dig...` into a variable

PR #302, opened 2022-03-16
DEV: Use reserved assigned notification type

xfalcox:


romanrizzi (commented):
Does that notification type exists? I don't see it listed in `Notification#types`

xfalcox:
> Does that notification type exists? I don't see it listed in `Notification#types`

https://github.com/discourse/discourse/pull/16207

romanrizzi (approved):

PR #306, opened 2022-03-17
FIX: Add missing notification title

xfalcox:


tshenry:
Nice! That should fix the title, but I'm pretty sure the icon will still be missing after this?

Screen Shot 2022-03-17 at 3 28 36 PM

I'm not sure exactly how it's defined, but this might be relevant:

https://github.com/discourse/discourse-assign/blob/264620985e6c7313510274c5136a5874b54b0346/assets/javascripts/discourse-assign/initializers/extend-for-assigns.js#L840-L848

tgxworld:
Is there a test we can add for this? There has already been multiple fixes with regards to missing translation strings so I think a test here will help greatly in catching regresson.

xfalcox:
If we had a way to test for any missing i18n strings it would be great. Not
sure how.

tgxworld:
Sorry I didn't mean we should test missing i18n strings. I think if we add an acceptance test, we would be able to ensure that all the strings that are supposed to be displayed have been displayed.

tshenry (approved):

PR #317, opened 2022-04-19
DRAFT: Add priorities to assignment

nattsw:
⚠️ **(page refresh at 0:06)**

https://user-images.githubusercontent.com/1555215/164071350-6c1db26b-0266-4f3f-9f23-e5adae36fc95.mov

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -8,6 +8,13 @@ class Assignment < ActiveRecord::Base
   belongs_to :assigned_by_user, class_name: "User"
   belongs_to :target, polymorphic: true
 
+  enum priority: {
+    low: 4,
+    medium: 3,
+    high: 2,
+    urgent: 1,
+  }, _prefix: true
+
   scope :joins_with_topics, -> { joins("INNER JOIN topics ON topics.id = assignments.target_id AND assignments.target_type = 'Topic' AND topics.deleted_at IS NULL") }
 
   def self.valid_type?(type)
@@ -37,6 +44,7 @@ end
 #  target_id           :integer          not null
 #  target_type         :string           not null
 #  active              :boolean          default(TRUE)
+#  priority            :integer
 #
 # Indexes
 #
--- a/assets/javascripts/discourse-assign/controllers/assign-user.js
+++ b/assets/javascripts/discourse-assign/controllers/assign-user.js
@@ -1,10 +1,18 @@
 import Controller, { inject as controller } from "@ember/controller";
+import { action } from "@ember/object";
+import { not, or } from "@ember/object/computed";
 import { inject as service } from "@ember/service";
+import { isEmpty } from "@ember/utils";
 import { ajax } from "discourse/lib/ajax";
 import { popupAjaxError } from "discourse/lib/ajax-error";
-import { not, or } from "@ember/object/computed";
-import { isEmpty } from "@ember/utils";
-import { action } from "@ember/object";
+import I18n from "I18n";
+
+const PRIORITIES = [
+  { name: I18n.t("discourse_assign.priorities.low"), value: 4 },
+  { name: I18n.t("discourse_assign.priorities.medium"), value: 3 },
+  { name: I18n.t("discourse_assign.priorities.high"), value: 2 },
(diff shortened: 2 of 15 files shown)


spec/serializers/topic_view_serializer_spec.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
```
tgxworld, line 3:
```suggestion
```

Minor but this is no longer required going forward since we've added this to `.rspec`

spec/requests/assign_controller_spec.rb
```diff
@@ -116,6 +116,15 @@
       expect(post.topic.reload.assignment.assigned_to_id).to eq(user2.id)
     end
 
+    it 'assigns topic with priority to a user' do
```
tgxworld, line 119: Can we also add a test case which passes in an invalid priority like a random string or an invalid number? My hunch is that this is not handled right now and might result in us writing invalid values into the DB.

app/models/assignment.rb
```diff
@@ -8,6 +8,13 @@ class Assignment < ActiveRecord::Base
   belongs_to :assigned_by_user, class_name: "User"
   belongs_to :target, polymorphic: true
 
+  enum priority: {
```
tgxworld, line 11: I'm not familiar with `enum` but does it add a validation to ensure that the value being written into the DB is within the list below? Kind of related to my comment in https://github.com/discourse/discourse-assign/pull/317/files#r853892117.

tgxworld:
Sorry my bad just realized this is a draft :grin:

spec/requests/assign_controller_spec.rb
nattsw: The DB actually is smart about it because of the enums declared. Let me see if I can get you an error...

nattsw: > ArgumentError: '5' is not a valid priority

app/models/assignment.rb
nattsw: (see above comment)

spec/requests/assign_controller_spec.rb
tgxworld: I do not think we have any special rescues for `ArgumentError` at the controller level so this might result in a 500 response. I think we might want to ensure that the priority passed in by the client is valid and return a nicer response.

assets/javascripts/discourse-assign/controllers/assign-user.js
```diff
@@ -13,6 +21,7 @@ export default Controller.extend({
   taskActions: service(),
   autofocus: not("capabilities.touch"),
   assigneeName: or("model.username", "model.group_name"),
+  priorities: PRIORITIES,
```
jjaffeux, line 24: you should set this in init, and have it as `priorities: null` here, this is mostly due to controllers being singletons in Ember.

jjaffeux:
> ⚠️ **(page refresh at 0:06)**
> 
>  Screen.Recording.2022-04-20.at.2.26.35.AM.mov

Where is priority displayed? don't see it in the video.

assets/javascripts/discourse-assign/controllers/assign-user.js
nattsw: I can see why "assignments" (`this.a = b`) may need to be done on `init`, but should this be the case even if `priorities` should be immutable -- they don't change at all nor by any setting -- and should be the same wherever invoked?

nattsw:
@jjaffeux Currently it is only displayed in the modal as shown in the video.

spec/requests/assign_controller_spec.rb
nattsw: Your concern that we're writing invalid values to the DB is addressed...

But for this one where it results in 500, I'm not sure if we have to care: the only consumer for this endpoint is our frontend, and the frontend only ever passes 1-4. It also feels a bit _off_ to test for the framework since this is a built-in enum feature.

davidtaylorhq:
Plugin moving to core

PR #319, opened 2022-04-26
FIX: Hide footer action button when user cannot assign

nattsw:
Fixes https://meta.discourse.org/t/public-assigns-shows-reassign-button/224860

> When “public assigns” is enabled in the assign plugin, a regular user sees a drop down button which looks like they are able to reassign the topic. The button does not actually work though.

### DIFF
--- a/test/javascripts/acceptance/assigned-topic-test.js
+++ b/test/javascripts/acceptance/assigned-topic-test.js
@@ -78,6 +78,7 @@ acceptance("Discourse Assign | Assigned topic", function (needs) {
     assign_enabled: true,
     tagging_enabled: true,
     assigns_user_url_path: "/",
+    assigns_public: true,
   });
 
   assignCurrentUserToTopic(needs);
@@ -124,7 +125,7 @@ acceptance("Discourse Assign | Assigned topic", function (needs) {
     );
   });
 
-  test("Public user cannot see footer button", async (assert) => {
+  test("User without assign ability cannot see footer button", async (assert) => {
     updateCurrentUser({ can_assign: false, admin: false, moderator: false });
     await visit("/t/assignment-topic/45");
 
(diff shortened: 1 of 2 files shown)


test/javascripts/acceptance/assigned-topic-test.js
```diff
@@ -123,6 +123,16 @@ acceptance("Discourse Assign | Assigned topic", function (needs) {
...
     );
   });
+
+  test("Public user cannot see footer button", async (assert) => {
+    updateCurrentUser({ can_assign: false, admin: false, moderator: false });
+    await visit("/t/assignment-topic/45");
+
+    assert.notOk(
+      exists("#topic-footer-dropdown-reassign"),
+      "shows reassign dropdown at the bottom of the topic"
```
tgxworld, line 133:
```suggestion
      "does not show reassign dropdown at the bottom of the topic"
```

test/javascripts/acceptance/assigned-topic-test.js
```diff
@@ -123,6 +123,16 @@ acceptance("Discourse Assign | Assigned topic", function (needs) {
       "shows reassign dropdown at the bottom of the topic"
     );
   });
+
+  test("Public user cannot see footer button", async (assert) => {
```
tgxworld, line 127: `Public` is a little vague and is not a common term that we use here. Perhaps something like `Footer button not displayed for user without the ability to assign`?

tgxworld (approved):
Aside from some minor comments, fix looks good to me.

test/javascripts/acceptance/assigned-topic-test.js
nattsw: I only used `Public` because the user setting says "public assigns" and the hint is "Allow general public to see..."

test/javascripts/acceptance/assigned-topic-test.js
nattsw: D'oh!!

PR #326, opened 2022-04-29
FEATURE: Add assign note

nattsw:
Add assign note to assignment /64635

Subsequent PR https://github.com/discourse/discourse-assign/pull/327 based on this branch shows assignment note in tooltip.

https://user-images.githubusercontent.com/1555215/166009891-746120ea-5dd3-46f1-b642-ab0b54006249.mov

### DIFF
--- a/app/controllers/discourse_assign/assign_controller.rb
+++ b/app/controllers/discourse_assign/assign_controller.rb
@@ -48,7 +48,7 @@ module DiscourseAssign
       target_type = params.require(:target_type)
       username = params.permit(:username)['username']
       group_name = params.permit(:group_name)['group_name']
-      note = params.permit(:note)['note']
+      note = params.permit(:note)['note'].presence
 
       assign_to = username.present? ? User.find_by(username_lower: username.downcase) : Group.where("LOWER(name) = ?", group_name.downcase).first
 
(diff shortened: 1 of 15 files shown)


app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -48,6 +48,7 @@ def assign
       target_type = params.require(:target_type)
       username = params.permit(:username)['username']
       group_name = params.permit(:group_name)['group_name']
+      note = params.permit(:note)['note']
```
CvX, line 51:
```suggestion
      note = params.permit(:note)['note'].presence
```

Just to avoid having both null and blank notes.

You could even go with:

```suggestion
      note = params.permit(:note)['note'].presence&.strip
```

to clean up the input a bit

CvX: Alternatively, this could be handled in `Assigner` class 🤷

CvX (approved):

app/controllers/discourse_assign/assign_controller.rb
nattsw: Good point. Keeping it in the controller as I prefer sanitising params there.

nattsw:
Thank you

PR #327, opened 2022-04-29
FEATURE: Show note in tooltip

nattsw:
To be merged after https://github.com/discourse/discourse-assign/pull/326. Related /64635

https://user-images.githubusercontent.com/1555215/166009725-0b73ce05-207d-4b55-ba31-ad7eb87bcd10.mov

### DIFF
--- a/assets/javascripts/discourse-assign/initializers/extend-for-assigns.js
+++ b/assets/javascripts/discourse-assign/initializers/extend-for-assigns.js
@@ -586,9 +586,18 @@ function initialize(api) {
   api.addDiscoveryQueryParam("assigned", { replace: true, refreshModel: true });
 
   api.addTagsHtmlCallback((topic, params = {}) => {
-    const [assignedToUser, assignedToGroup] = Object.values(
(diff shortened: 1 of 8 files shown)


assets/javascripts/discourse-assign/initializers/extend-for-assigns.js
```diff
@@ -629,7 +639,7 @@ function initialize(api) {
               : "";
           return `<${tagName} class="assigned-to discourse-tag simple" ${href}>
         ${icon}
-        <span>${name}</span>
+        <span title="${note}">${name}</span>
```
tgxworld, line 642: I didn't confirm this but I think we have to escape `note` here because it is user input.

tgxworld (approved):

assets/javascripts/discourse-assign/initializers/extend-for-assigns.js
nattsw: Good point thanks!

PR #329, opened 2022-05-10
FIX: Assigning a user without notes assigns and closes the modal

nattsw:
Quite literally it.

Previously, clicking on the user suggestion immediately triggered the assign and closes the modal. It was reverted (https://github.com/discourse/discourse-assign/pull/326) because there was an additional note box to fill in. But folks prefer immediately assigning despite notes not filled 🤷‍♀️

### DIFF
--- a/test/javascripts/controllers/assign-user-test.js
+++ b/test/javascripts/controllers/assign-user-test.js
@@ -1,28 +1,55 @@
+import EmberObject from "@ember/object";
+import pretender from "discourse/tests/helpers/create-pretender";
 import { discourseModule } from "discourse/tests/helpers/qunit-helpers";
 import { test } from "qunit";
-import pretender from "discourse/tests/helpers/create-pretender";
-import EmberObject from "@ember/object";
 
 discourseModule("Unit | Controller | assign-user", function (hooks) {
-  hooks.beforeEach(function () {
+  hooks.beforeEach(function () {});
+
+  test("doesn't set suggestions and fails gracefully if controller is destroyed", function (assert) {
+    let resolveSuggestions;
+    pretender.get("/assign/suggestions", () => {
+      // eslint-disable-next-line no-restricted-globals
+      return new Promise((resolve) => {
+        resolveSuggestions = resolve;
+      });
+    });
+    const controller = this.getController("assign-user", {
+      model: {
+        target: EmberObject.create({}),
+      },
+    });
+
+    controller.destroy();
+    resolveSuggestions([
+      200,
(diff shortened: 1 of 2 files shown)


CvX:
Dunno if the failure is related to this change:

`calling set on destroyed object: <(unknown):ember1456>.assignSuggestions = `

but isDestroying || isDestroyed should be added somewhere either way 😛

nattsw:
I don't think it's related to this change -- this change is in the assign modal, whereas the failure is in `/u/eviltrout/preferences/notifications`... 🤔

test/javascripts/controllers/assign-user-test.js
```diff
@@ -0,0 +1,61 @@
...
+      },
+    ]);
+    const controller = this.getController("assign-user", {
+      model: {
+        target: EmberObject.create({}),
+      },
+    });
+    controller.destroy();
+
+    assert.strictEqual(controller.get("assign_allowed_on_groups"), undefined);
```
nattsw, line 26: The absence of `if (this.isDestroying || this.isDestroyed)` will cause this to fail, obv.

nattsw: OHHH we can

test/javascripts/controllers/assign-user-test.js
```diff
@@ -0,0 +1,69 @@
...
+import { test } from "qunit";
+
+discourseModule("Unit | Controller | assign-user", function (hooks) {
+  hooks.beforeEach(function () {});
+
+  test("doesn't set suggestions and fails gracefully if controller is destroyed", function (assert) {
+    let resolveSuggestions;
+    pretender.get("/assign/suggestions", () => {
+      // eslint-disable-next-line no-restricted-globals
+      return new Promise((resolve) => {
```
CvX, line 13: For now it's still recommended to use `import { Promise } from "rsvp";`

At some point we'll drop rsvp dependency but for now it's still standard in the codebase

test/javascripts/controllers/assign-user-test.js
```diff
@@ -0,0 +1,69 @@
+import EmberObject from "@ember/object";
+import pretender from "discourse/tests/helpers/create-pretender";
+import { discourseModule } from "discourse/tests/helpers/qunit-helpers";
+import { test } from "qunit";
+
+discourseModule("Unit | Controller | assign-user", function (hooks) {
+  hooks.beforeEach(function () {});
+
```
CvX, line 8:
```suggestion
```

CvX (approved):

PR #334, opened 2022-05-12
FIX: Only assign when suggestion is clicked

nattsw:
Screenshot 2022-05-12 at 5 17 08 PM

### DIFF
--- a/test/javascripts/controllers/assign-user-test.js
+++ b/test/javascripts/controllers/assign-user-test.js
@@ -2,37 +2,9 @@ import EmberObject from "@ember/object";
 import pretender from "discourse/tests/helpers/create-pretender";
 import { discourseModule } from "discourse/tests/helpers/qunit-helpers";
 import { test } from "qunit";
-import { Promise } from "rsvp";
 
 discourseModule("Unit | Controller | assign-user", function () {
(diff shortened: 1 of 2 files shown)


nattsw:
Ah

`Assertion Failed: calling set on destroyed object: <(unknown):ember10389>.model = [object Object]`

test/javascripts/controllers/assign-user-test.js
```diff
@@ -2,37 +2,42 @@ import EmberObject from "@ember/object";
...
-    controller.destroy();
-    resolveSuggestions([
-      200,
-      { "Content-Type": "application/json" },
-      {
-        suggestions: [],
-        assign_allowed_on_groups: ["nat"],
-        assign_allowed_for_groups: [],
-      },
-    ]);
```
nattsw, line 30: Turns out I have to delete this `isDestroying || isDestroyed` test because there's no good way of simulating a destroyed controller. Ember does not allow me to `set("isDestroyed")` nor recreate the controller after I destroy it.

The only reason it was passing previously was due to the seed which made this run after the subsequent test. With other seeds the controller gets destroyed first then every other test will fail. 😓

CvX (approved):

nattsw:
Thank you @CvX

PR #335, opened 2022-05-12
FEATURE: Shows note in moderator post

nattsw:
Shows the assignment note in moderator post.

Note there were no migrations to retroactively add the notes into small action posts
- A small action post generated from assignment has no relation to its topic/post
  - its `PostCustomField`s used in assign are used elsewhere as well (e.g. `action_code_who` etc)
- Vice versa, an assignment does not have relationship to its generated small action post

Each note modification also adds a moderator post, rather than modify the previous post. If we wish to have that, it's possible too.

Screenshot 2022-05-12 at 10 21 14 PM


nattsw:
Thank you @oblakeerickson

PR #346, opened 2022-06-07
UX: don't mention the group when no one is assigned.

vinothkannans:
It's a minor copyedit to remove group mention from the post raw. Else it sends a notification to all group members when no one is found to be assigned.


pmusaraj:
@vinothkannans spec failures look relevant, sadly.

PR #349, opened 2022-06-15
FIX: ensure that assignee is participant of pm

lis2:
We are already checking that assignee has access to the private message. However, admin still can be assigned as technically they have access.

We should ensure that assignee has direct access to the message.

### DIFF
--- a/lib/assigner.rb
+++ b/lib/assigner.rb
@@ -160,7 +160,12 @@ class ::Assigner
     @post_target ||= @target.is_a?(Post)
   end
 
+  def private_message_allowed_user_ids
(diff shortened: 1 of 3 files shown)


lib/assigner.rb
```diff
@@ -160,7 +160,12 @@ def post_target?
     @post_target ||= @target.is_a?(Post)
   end
 
+  def private_message_allowed_user_ids
+    @private_message_allowed_user_ids ||= topic.all_allowed_users.map(&:id)
```
CvX, line 164:
```suggestion
    @private_message_allowed_user_ids ||= topic.all_allowed_users.pluck(:id)
```

CvX (approved):

PR #351, opened 2022-06-17
FIX: Don't reuse assign note when reassigning

nattsw:
https://user-images.githubusercontent.com/1555215/174345166-b92c3f4a-13c7-4fd0-a3b7-7f0b05af2cef.mov

### DIFF
--- a/assets/javascripts/discourse/services/task-actions.js
+++ b/assets/javascripts/discourse/services/task-actions.js
@@ -49,7 +49,6 @@ export default Service.extend({
         username: user.username,
         target_id: target.id,
         target_type: targetType,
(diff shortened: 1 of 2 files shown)


pmusaraj (approved):
Looks good! 

Stupid question: does the unit test test that the note isn't used when reassigning?

nattsw:
> Stupid question: does the unit test test that the note isn't used when reassigning?

OMG!!

You totally caught the other part that the note was still there!

https://github.com/discourse/discourse-assign/blob/169f13e2396db5306558d5a405e38529ab257a94/assets/javascripts/discourse/services/task-actions.js#L53

PR #356, opened 2022-07-05
UX: Don't display assign user menu glyph when sidebar is enabled

tgxworld:
Assignments can already be accessed via sidebar.

### Before

Screenshot from 2022-07-05 15-42-12

### After

Screenshot from 2022-07-05 15-42-19

### DIFF
--- a/assets/javascripts/discourse-assign/initializers/extend-for-assigns.js
+++ b/assets/javascripts/discourse-assign/initializers/extend-for-assigns.js
@@ -649,7 +649,11 @@ function initialize(api) {
   });
 
   api.addUserMenuGlyph((widget) => {
-    if (widget.currentUser && widget.currentUser.can_assign) {
+    if (
+      widget.currentUser &&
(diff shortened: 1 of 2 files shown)


nattsw (approved):

assets/javascripts/discourse-assign/initializers/extend-for-assigns.js
```diff
@@ -649,7 +649,11 @@ function initialize(api) {
   });
 
   api.addUserMenuGlyph((widget) => {
-    if (widget.currentUser && widget.currentUser.can_assign) {
+    if (
+      widget.currentUser &&
+      widget.currentUser.can_assign &&
+      !widget.currentUser.experimental_sidebar_enabled
```
nattsw, line 655: `!undefined` is true, so this should be okay.

tgxworld:
Thank you for reviewing @nattsw

PR #363, opened 2022-07-29
FEATURE: Assign Status

xfalcox:
Adds a new plugin setting that when enabled adds a status field for every assignment. This setting defaults to off.

The possible status for an assignment are customizable via yet another new setting, and the first one on this list will be the default status for new assignments.

The status is not yet show anywhere except the assign modal and the small action posts in topics at the moment. Adding status to the assignment list for users and groups will be handled in the near future.

image
image
image

### DIFF
--- a/assets/javascripts/discourse-assign/controllers/assign-user.js
+++ b/assets/javascripts/discourse-assign/controllers/assign-user.js
@@ -1,6 +1,7 @@
 import Controller, { inject as controller } from "@ember/controller";
 import ModalFunctionality from "discourse/mixins/modal-functionality";
 import { action } from "@ember/object";
+import discourseComputed from "discourse-common/utils/decorators";
 import { not, or } from "@ember/object/computed";
 import { inject as service } from "@ember/service";
 import { isEmpty } from "@ember/utils";
@@ -40,6 +41,11 @@ export default Controller.extend(ModalFunctionality, {
       type: "assign",
       username,
     });
+  },
+
+  @discourseComputed("siteSettings.assign_statuses")
+  availiableStatuses() {
+    return this.siteSettings.assign_statuses.split('|').map((status) => { return { id: status, name: status }; });
   },
 
   @action
@@ -82,6 +88,7 @@ export default Controller.extend(ModalFunctionality, {
         target_id: this.get("model.target.id"),
         target_type: this.get("model.targetType"),
         note: this.get("model.note"),
+        status: this.get("model.status"),
       },
     })
       .then(() => {
--- a/assets/javascripts/discourse-assign/controllers/assign-user.js
+++ b/assets/javascripts/discourse-assign/controllers/assign-user.js
@@ -15,7 +15,6 @@ export default Controller.extend(ModalFunctionality, {
   taskActions: service(),
   autofocus: not("capabilities.touch"),
   assigneeName: or("model.username", "model.group_name"),
-  status: or("model.status", "model.target.assignment_status"),
 
   init() {
     this._super(...arguments);
@@ -54,6 +53,15 @@ export default Controller.extend(ModalFunctionality, {
     return this.siteSettings.assign_statuses.split("|").map((status) => {
       return { id: status, name: status };
     });
+  },
+
+  @discourseComputed("siteSettings.assign_statuses", "model.status")
+  status() {
+    return (
+      this.model.status ||
+      this.model.target.assignment_status ||
+      this.siteSettings.assign_statuses.split("|")[0]
+    );
   },
 
   @action
--- a/assets/javascripts/discourse-assign/controllers/assign-user.js
+++ b/assets/javascripts/discourse-assign/controllers/assign-user.js
@@ -15,6 +15,7 @@ export default Controller.extend(ModalFunctionality, {
   taskActions: service(),
   autofocus: not("capabilities.touch"),
   assigneeName: or("model.username", "model.group_name"),
+  status: or("model.status", "model.target.assignment_status"),
 
   init() {
     this._super(...arguments);
--- a/assets/javascripts/discourse-assign/controllers/assign-user.js
+++ b/assets/javascripts/discourse-assign/controllers/assign-user.js
@@ -49,7 +49,7 @@ export default Controller.extend(ModalFunctionality, {
   },
 
   @discourseComputed("siteSettings.assign_statuses")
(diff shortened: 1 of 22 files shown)


config/locales/server.en.yml
```diff
@@ -15,6 +15,13 @@ en:
     remind_assigns_frequency: "Frequency for reminding users about assigned topics."
     max_assigned_topics: "Maximum number of topics that can be assigned to a user."
     assign_allowed_on_groups: "Users in these groups are allowed to assign topics and can be assigned topics."
+    enable_assign_status: "Add a customizable status field to every assignment."
+    assign_statuses: "List of the possible status an assignment can have. First one is the default status applied to every new assignment."
```
pmusaraj, line 19:
```suggestion
    assign_statuses: "List of statuses available to each assignment. The first status is the default status applied to every new assignment."
```

config/locales/server.en.yml
```diff
@@ -15,6 +15,13 @@ en:
...
+    errors:
+      assign_statuses:
+        too_few: "Minimum of two different status required"
+        duplicate: "There are duplicated status values"
```
pmusaraj, line 23:
```suggestion
        duplicate: "There are duplicate status values."
```

config/locales/server.en.yml
```diff
@@ -15,6 +15,13 @@ en:
     remind_assigns_frequency: "Frequency for reminding users about assigned topics."
     max_assigned_topics: "Maximum number of topics that can be assigned to a user."
     assign_allowed_on_groups: "Users in these groups are allowed to assign topics and can be assigned topics."
+    enable_assign_status: "Add a customizable status field to every assignment."
+    assign_statuses: "List of the possible status an assignment can have. First one is the default status applied to every new assignment."
+    errors:
+      assign_statuses:
+        too_few: "Minimum of two different status required"
```
pmusaraj, line 22:
```suggestion
        too_few: "There must be at least two different statuses available."
```

config/locales/server.en.yml
```diff
@@ -77,3 +84,10 @@ en:
   discourse_push_notifications:
     popup:
       assigned: "@%{username} assigned you"
+  activerecord:
+    errors:
+      models:
+        assignment:
+          attributes:
+            status:
+              invalid: "Status is not on the possbile status list."
```
pmusaraj, line 93:
```suggestion
              invalid: "Selected status is invalid (it is not included in the assigned_status site setting)."
```

spec/requests/assign_controller_spec.rb
```diff
@@ -4,7 +4,10 @@
 require_relative "../support/assign_allowed_group"
 
 RSpec.describe DiscourseAssign::AssignController do
-  before { SiteSetting.assign_enabled = true }
+  before do
+    SiteSetting.assign_enabled = true
+    SiteSetting.enable_assign_status = true
```
pmusaraj, line 9: Can we move this to a context so that it is not enabled for all the existing specs, but rather only for the new specs that are specific to the status feature?

spec/serializers/post_serializer_spec.rb
```diff
@@ -39,4 +40,10 @@
     serializer = PostSerializer.new(post, scope: guardian)
     expect(serializer.as_json[:post][:assignment_note]).to eq("tomtom best")
   end
+
+  it "includes status in serializer" do
```
pmusaraj, line 44: Would be good to add a spec here that checks that status is not included when `enable_assign_status` is not enabled.

test/javascripts/acceptance/assigned-topic-test.js
```diff
@@ -89,6 +90,7 @@ acceptance("Discourse Assign | Assigned topic", function (needs) {
     tagging_enabled: true,
     assigns_user_url_path: "/",
     assigns_public: true,
+    enable_assign_status: true,
```
pmusaraj, line 93: We should have tests that check the status dropdown is present (or not) depending on when this setting is enabled or not, right?

assets/javascripts/discourse-assign/controllers/assign-user.js
```diff
@@ -42,6 +43,18 @@ export default Controller.extend(ModalFunctionality, {
     });
   },
 
+  @discourseComputed("siteSettings.assign_statuses")
+  statusEnabled() {
+    return this.siteSettings.enable_assign_status;
+  },
+
+  @discourseComputed("siteSettings.assign_statuses")
+  availiableStatuses() {
```
pmusaraj, line 52:
```suggestion
  availableStatuses() {
```

assets/javascripts/discourse/templates/modal/assign-user.hbs
```diff
@@ -25,8 +25,21 @@
...
       {{/each}}
     </div>
-    <label>{{i18n "discourse_assign.assign_modal.note_label"}}</label>
-    {{textarea id="assign-modal-note" value=model.note key-down=(action "handleTextAreaKeydown")}}
+    {{#if this.statusEnabled}}
+      <div class="control-group assign-status">
+        <label>{{i18n "discourse_assign.assign_modal.status_label"}}</label>
+        {{combo-box
+          id="assign-status"
+          content=availiableStatuses
```
pmusaraj, line 33:
```suggestion
          content=availableStatuses
```

assets/javascripts/discourse/templates/modal/assign-user.hbs
```diff
@@ -25,8 +25,21 @@
...
+          value=model.status
```
pmusaraj, line 34: This means that the default value of the dropdown is empty: 

image

where in fact it should be the default status (i.e. New), right?

pmusaraj (commented):
Looks generally good, it's mostly small comments and typo fixes. 

One thing though that would be good is to use the FEATURE: prefix in the PR name and add a description. It helps people using the plugin be aware of the new feature and also highlights this for the next release notes (otherwise, it will be missed).

assets/javascripts/discourse/templates/modal/assign-user.hbs
xfalcox: We are setting the first status as a default in the pre-validation model hook too, but I guess it would be better UX to set it as default in the UI too?

pmusaraj (approved):
Changes look great, thanks!

PR #366, opened 2022-08-12
FEATURE: Add assigns tab to the experimental user menu

OsamaSayegh:
This PR adds a tab for assignments in the experimental user menu. Screenshot:

The assignments tab behaves very similarly to the bookmarks and messages tab in core: it displays the user's unread assign notifications first and then fills the rest of available space in the menu with assignments (same content that the current user menu displays).

Core PR https://github.com/discourse/discourse/pull/17879 is required before this can be merged.

More details of the experimental user menu can be found in https://github.com/discourse/discourse/pull/17379.


app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -174,6 +174,50 @@ def group_members
              }
     end
 
+    USER_MENU_LIMIT = 20
```
tgxworld, line 177: Are we able to pull this constant from core instead of defining it here? It'll at least help to ensure any changes in core is automatically reflected in the plugin.

assets/javascripts/discourse/components/user-menu/assigns-list.js
```diff
@@ -0,0 +1,78 @@
...
+    modalController.set(
+      "confirmationMessage",
+      I18n.t("notifications.dismiss_confirmation.body.assigns", {
+        count: this._unreadAssignedNotificationsCount,
+      })
+    );
+    return modalController;
+  }
+
+  get _unreadAssignedNotificationsCount() {
```
tgxworld, line 70:
```suggestion
  get #unreadAssignedNotificationsCount() {
```

assets/javascripts/discourse/initializers/assign-user-menu.js
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        const currentUser = api.getCurrentUser();
+        if (!currentUser?.can_assign) {
+          return;
+        }
+        api.registerUserMenuTab((UserMenuTab) => {
+          return class extends UserMenuTab {
+            get id() {
+              return "assign-notifications";
```
tgxworld, line 21:
```suggestion
              return "assign-list";
```

The list is not all notifications so using `notifications` here doesn't feel right.

spec/requests/assign_controller_spec.rb
```diff
@@ -392,4 +392,133 @@
...
+        user.update!(admin: false)
+        user.group_users.where(group_id: default_allowed_group.id).destroy_all
+        get "/assign/user-menu-assigns.json"
+        expect(response.status).to eq(403)
+      end
+
+      it "responds with 404 if the assign_enabled setting is disabled" do
+        SiteSetting.assign_enabled = false
+        get "/assign/user-menu-assigns.json"
+        expect(response.status).to eq(403)
```
tgxworld, line 474:
```suggestion
        expect(response.status).to eq(404)
```

The description of this test says it should return 404.

spec/requests/assign_controller_spec.rb
```diff
@@ -392,4 +392,133 @@
...
+        get "/assign/user-menu-assigns.json"
+        expect(response.status).to eq(200)
+
+        notifications = response.parsed_body["notifications"]
+        expect(notifications.map { |n| [n["topic_id"], n["post_number"]] }).to contain_exactly(
+          [unread_assigned_topic.id, 1],
+          [unread_assigned_post.topic.id, unread_assigned_post.post_number],
+          [unread_assigned_post_in_same_topic.topic.id, unread_assigned_post_in_same_topic.post_number]
+        )
+      end
```
tgxworld, line 487: Minor note but `contain_exactly` ignores order which I think is important here.

assets/javascripts/discourse/components/user-menu/assigns-list.js
OsamaSayegh: I did do this initially, but I got an error that we couldn't use this feature in plugins yet.

app/controllers/discourse_assign/assign_controller.rb
OsamaSayegh: done!

tgxworld (approved):

PR #368, opened 2022-08-16
FIX: Check if assignment has same user and details

nbianca:
The check existed, but its implementation was incorrect and it did not
work when the target was a post.

### DIFF
--- a/lib/assigner.rb
+++ b/lib/assigner.rb
@@ -211,9 +211,7 @@ class ::Assigner
       end
     when !can_be_assigned?(assign_to)
       assign_to.is_a?(User) ? :forbidden_assign_to : :forbidden_group_assign_to
-    when topic_same_assignee_and_details(assign_to, type, note, status)
-      assign_to.is_a?(User) ? :already_assigned : :group_already_assigned
-    when post_same_assignee_and_details(assign_to, type, note, status)
+    when already_assigned?(assign_to, type, note, status)
       assign_to.is_a?(User) ? :already_assigned : :group_already_assigned
     when Assignment.where(topic: topic).count >= ASSIGNMENTS_PER_TOPIC_LIMIT
       :too_many_assigns_for_topic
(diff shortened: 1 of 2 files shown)


lib/assigner.rb
```diff
@@ -523,7 +523,15 @@ def topic_same_assignee_and_details(assign_to, type, note, status)
   end
 
   def post_same_assignee_and_details(assign_to, type, note, status)
-    @target.is_a?(Topic) &&
+    if @target.is_a?(Post)
+      @target.assignment&.assigned_to_id == assign_to.id &&
+      @target.assignment&.assigned_to_type == type && @target.assignment.active == true &&
```
ZogStriP, line 528:
```suggestion
      @target.assignment&.assigned_to_type == type && @target.assignment.active &&
```

lib/assigner.rb
```diff
@@ -523,7 +523,15 @@ def topic_same_assignee_and_details(assign_to, type, note, status)
...
+      @target.assignment&.note == note &&
+      (
+        @target.assignment&.status == status ||
+          @target.assignment&.status == Assignment.default_status && status.nil?
```
ZogStriP, line 532: Should this be inside a parenthesis? Otherwise it's hard to know if `||` has precedence over `&&`

```suggestion
          (@target.assignment&.status == Assignment.default_status && status.nil?)
```

ZogStriP (approved):

lib/assigner.rb
nbianca: I agree it can be confusing at first. I changed the other places too.

lib/assigner.rb
```diff
@@ -514,26 +514,35 @@ def moderator_post_unassign_action_code(assignment)
 
   def topic_same_assignee_and_details(assign_to, type, note, status)
```
ZogStriP, line 515: Could we merge `topic_same_assignee_and_details` and `post_same_assignee_and_details` into just one method? Seems like there's a lot of repetition for no good reason.

lib/assigner.rb
```diff
@@ -211,9 +211,7 @@ def forbidden_reasons(assign_to:, type:, note:, status:)
       end
     when !can_be_assigned?(assign_to)
       assign_to.is_a?(User) ? :forbidden_assign_to : :forbidden_group_assign_to
-    when topic_same_assignee_and_details(assign_to, type, note, status)
-      assign_to.is_a?(User) ? :already_assigned : :group_already_assigned
-    when post_same_assignee_and_details(assign_to, type, note, status)
+    when post_or_topic_same_assignee_and_details(assign_to, type, note, status)
```
ZogStriP, line 214: That method name is a mouthful 🤔 Maybe `already_assigned?` would work better?

lib/assigner.rb
```diff
@@ -512,28 +510,38 @@ def moderator_post_unassign_action_code(assignment)
...
+    return if @target.assignment&.assigned_to_type != type
+    return if !@target.assignment&.active
+    return if @target.assignment&.note != note
+    if @target.assignment&.status != status &&
+         (@target.assignment&.status != Assignment.default_status || status.present?)
+      return
+    end
+
+    true
   end
```
ZogStriP, line 524: How about this?

```suggestion
  def target_same_assignee_and_details(assign_to, type, note, status)
    return if !@target.assignment&.active
    return if @target.assignment&.assigned_to_id != assign_to.id
    return if @target.assignment&.assigned_to_type != type
    return if @target.assignment&.note != note
    return if @target.assignment&.status != (status.presence || Assignment.default_status)
    true
  end
```

lib/assigner.rb
```diff
@@ -512,28 +510,38 @@ def moderator_post_unassign_action_code(assignment)
...
+    Assignment
+      .where(topic_id: topic.id, target_type: "Post", active: true)
+      .any? do |assignment|
+        next if assignment.assigned_to_id != assign_to.id
+        next if assignment.assigned_to_type != type
+        next if assignment.note != note
+        if assignment.status != status &&
+             (assignment.status != Assignment.default_status || status.present?)
+          next
         end
```
ZogStriP, line 536: This should work?

```suggestion
        next if assignment.status != (status.presence || Assignment.default_status)
```

lib/assigner.rb
```diff
@@ -512,28 +510,31 @@ def moderator_post_unassign_action_code(assignment)
...
-        end
+  def assignee_was_assigned_in_topic(assign_to, type, note, status)
+    Assignment
+      .where(topic_id: topic.id, target_type: "Post", active: true)
+      .any? do |assignment|
+        next if assignment.assigned_to_id != assign_to.id
+        next if assignment.assigned_to_type != type
+        next if assignment.note != note
+        next if assignment.status != (status.presence || Assignment.default_status)
+
```
ZogStriP, line 530: Do we need this empty line?

ZogStriP (approved):

lib/assigner.rb
nbianca: It's slightly different, but I just moved everything into a method because I can reuse it.

PR #370, opened 2022-08-23
FIX: Find better users for automatic assignment

nbianca:
There were two problems with the way current automation script for automatic assignment works:

* it tried to assign users that were not allowed to be assigned because they were not part of groups that are allowed to use discourse-assign - fixed by skipping users that are not a part of assign_allowed_on_groups groups

* it assigned new users - fixed by adding a new user that can skip new users

### DIFF
--- a/lib/random_assign_utils.rb
+++ b/lib/random_assign_utils.rb
@@ -38,7 +38,7 @@ class RandomAssignUtils
       raise_error(automation, "Group(#{group_id}) not found")
     end
 
-    allowed_assign_user_ids = User.assign_allowed.pluck(:id)
+    assignable_user_ids = User.assign_allowed.pluck(:id)
     users_on_holiday =
       Set.new(
         User.where(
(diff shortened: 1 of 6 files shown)


lib/random_assign_utils.rb
```diff
@@ -38,19 +38,23 @@ def self.automation_script!(context, fields, automation)
       raise_error(automation, "Group(#{group_id}) not found")
     end
 
+    allowed_assign_user_ids = User.assign_allowed.pluck(:id)
```
ZogStriP, line 41: Maybe a better name would be?

```suggestion
    assignable_user_ids = User.assign_allowed.pluck(:id)
```

spec/lib/random_assign_utils_spec.rb
```diff
@@ -78,6 +78,34 @@
       end
     end
 
+    context "no users of group can be assigned because are not members of assign_allowed_on_groups groups" do
```
ZogStriP, line 81:
```suggestion
    context "no users can be assigned because none are members of assign_allowed_on_groups groups" do
```

spec/lib/random_assign_utils_spec.rb
```diff
@@ -78,6 +78,34 @@
...
+              "value" => group_1.id,
+            },
+            "assigned_topic" => {
+              "value" => topic_1.id,
+            },
+          },
+          automation,
+        )
+        expect(topic_1.posts.first.raw).to match(
+          "Attempted randomly assign a member of `@#{group_1.name}`, but no one was available.",
```
ZogStriP, line 104: Will this break if we ever change the string?

nbianca: I fixed this test and all the other ones.

ZogStriP (approved):

PR #375, opened 2022-08-30
FIX: Count only active assignments when checking limits

nbianca:
Only 5 assignments can exist simultaneously for a topic or its post.
The code that counted the assignments did not ignore the inactive
assignments which made the limit check more strict than it should have
been.


ZogStriP (approved):

ZogStriP:
Love ❤️ the test 👏

PR #379, opened 2022-09-16
FIX: Don't fail because flair_upload wasn't eager loaded

romanrizzi:
The basic group serializer complains that the flair_upload relation has to be eager loaded when attempting to included assigned groups in search results, which is a consequence of using strict loading during search's preload to avoid N+1.

Eager loading the association during this would definitely fix the problem, but is it the right solution? It seems to me that this serializer does too much, when we only need a small number of attributes to display the assigned widget. Instead, this commit adds a lightweight custom serializer which grabs only the essentials for it to work.

### DIFF
--- a/app/serializers/assigned_group_serializer.rb
+++ b/app/serializers/assigned_group_serializer.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class AssignedGroupSerializer < ApplicationSerializer
+  attributes :id, :name, :assign_icon, :assign_path
+
+  def assign_icon
+    "group-plus"
+  end
+
+  def assign_path
+    "/g/#{object.name}/assigned/everyone"
+  end
+end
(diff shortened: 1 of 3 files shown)


app/serializers/assigned_group_serializer.rb
```diff
@@ -0,0 +1,13 @@
...
+
+class AssignedGroupSerializer < ApplicationSerializer
+  attributes :id, :name, :assign_icon, :assign_path
+
+  def assign_icon
+    "group-plus"
+  end
+
+  def assign_path
+    "/g/#{object.name}/assigned/everyone"
```
pmusaraj, line 11: This isn't new to this PR, but it doesn't look like this path is being used anywhere...

pmusaraj (approved):
Yes, a much more basic serializer makes sense here, good call.

app/serializers/assigned_group_serializer.rb
pmusaraj: This shouldn't hold merging the PR, but it does look like a separate regression.

romanrizzi: Will keep the attribute then. A lot has changed since the last time I worked on this plugin, so not sure how it should be used.

PR #380, opened 2022-09-17
DEV: Make the assign tab in the user menu link to the assign page

OsamaSayegh:
As of https://github.com/discourse/discourse/commit/12ebdf0ff0f742d5363527908a4a2721e41b00f2, it's possible to make tabs in the experimental user menu link to a page when they're clicked on when they're active. This PR makes the assign tab links to the `/my/activity/assigned` when it's active.

### DIFF
--- a/assets/javascripts/discourse/initializers/assign-user-menu.js
+++ b/assets/javascripts/discourse/initializers/assign-user-menu.js
@@ -36,6 +36,10 @@ export default {
             get notificationTypes() {
               return ["assigned"];
             }
+
+            get linkWhenActive() {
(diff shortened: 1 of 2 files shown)


assets/javascripts/discourse/initializers/assign-user-menu.js
```diff
@@ -36,6 +36,10 @@ export default {
             get notificationTypes() {
               return ["assigned"];
             }
+
+            get linkWhenActive() {
+              return `${this.currentUser.path}/activity/assigned`;
```
nattsw, line 41: Is there a simple test we can add?

pmusaraj (approved):

PR #390, opened 2022-10-21
DEV: Persist records to DB in tests

tgxworld:
There is almost no reason for us to use `Fabricate.build` which does not persist the record to the database and is not what we encounter in production.


CvX (approved):

tgxworld:
Thank you @CvX

PR #401, opened 2022-12-14
UX: update small actions to use sentence case

awesomerobot:
@discourse-translator-bot keep_translations_and_approvals

This updates small actions to use sentence case. Previously these were all-caps, which hid some casing inconsistency. 

Pairs with the core PR https://github.com/discourse/discourse/pull/19470, see some additional reasoning in that description.


pmusaraj (approved):

nbianca (approved):

CvX:
@gschlager will the translators know that these need to be updated? 🤔

gschlager:
> @gschlager will the translators know that these need to be updated?

No, not really, but it should be fine. Most translators seem to have used the correct uppercase anyway. I was more concerned about losing existing translations than about having some inconsistencies.

PR #418, opened 2023-01-03
DEV: Change flaky spec to provide more information when it fails

tgxworld:
This spec has been detected to be flaky by
https://github.com/discourse/flaky-tests-detective which we run
internally. However, I could not reproduce the flakiness after multiple
attempts so I rewrote the assertions such that we set up the
expectations of what `Assignment#assigned_to_id` should be before and
after the triggering of the DiscourseEvent.


lis2 (approved):

tgxworld:
Thank you @lis2

PR #419, opened 2023-01-05
FIX: reassign when assignment limit is fulfilled

lis2:
Previously, when topic was already assigned 5 times, reassign was not possible. User had to first unassign topic and then assign to someone else.

### DIFF
--- a/lib/assigner.rb
+++ b/lib/assigner.rb
@@ -213,7 +213,8 @@ class ::Assigner
       assign_to.is_a?(User) ? :forbidden_assign_to : :forbidden_group_assign_to
     when already_assigned?(assign_to, type, note, status)
       assign_to.is_a?(User) ? :already_assigned : :group_already_assigned
-    when Assignment.where(topic: topic, active: true).count >= ASSIGNMENTS_PER_TOPIC_LIMIT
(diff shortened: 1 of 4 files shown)


pmusaraj:
@jancernik did you want to review this since you're spending time making fixes to the plugin?

lib/assigner.rb
```diff
@@ -213,7 +213,8 @@ def forbidden_reasons(assign_to:, type:, note:, status:)
       assign_to.is_a?(User) ? :forbidden_assign_to : :forbidden_group_assign_to
     when already_assigned?(assign_to, type, note, status)
       assign_to.is_a?(User) ? :already_assigned : :group_already_assigned
-    when Assignment.where(topic: topic, active: true).count >= ASSIGNMENTS_PER_TOPIC_LIMIT
```
Drenmi, line 216: Out of scope for this PR, but seeing this line (the count in particular) made me curious.

Screenshot 2023-01-12 at 11 19 52 AM

Maybe we should have a counter cache for this. At the very least we should probably add an index on `topic_id` so we don't have to walk the entire `assignments` table. Maybe we create a #todo for this, WDYT? 🙂

lis2: Excellent catch! I added migration to add topic_id index

jancernik (approved):

PR #425, opened 2023-01-11
FIX: Broken posts when assigning a closed topic

jancernik:
1. Assign a topic to someone
2. Close the topic with the setting unassign_on_close enabled
3. Assign the topic to the same user as before

This caused small broken or empty posts to be displayed.

### DIFF
--- a/spec/system/assign_topic_spec.rb
+++ b/spec/system/assign_topic_spec.rb
@@ -1,0 +1,105 @@
+# frozen_string_literal: true
+
+describe "Assign | Assigning topics", type: :system, js: true do
+  let(:topic_page) { PageObjects::Pages::Topic.new }
+  let(:assign_modal) { PageObjects::Modals::Assign.new }
+  fab!(:staff_user) { Fabricate(:user, groups: [Group[:staff]]) }
+  fab!(:admin) { Fabricate(:admin) }
+  fab!(:topic) { Fabricate(:topic) }
+  fab!(:post) { Fabricate(:post, topic: topic) }
+
+  before do
+    SiteSetting.assign_enabled = true
+    sign_in(admin)
+  end
+
+  describe "with open topic" do
+    it "can assign and unassign" do
+      visit "/t/#{topic.id}"
+
+      topic_page.click_assign_topic
+      assign_modal.set_assignee(staff_user)
+      assign_modal.confirm
+
+      expect(find("#post_2")).to have_content("Assigned")
+      expect(find("#topic .assigned-to")).to have_content(staff_user.username)
+
+      topic_page.click_unassign_topic
+
+      expect(find("#post_3")).to have_content("Unassigned")
+      expect(page).not_to have_css("#topic .assigned-to")
+    end
+
+    context "when unassign_on_close is set to true" do
+      before { SiteSetting.unassign_on_close = true }
+
+      it "topic gets unassigned on close" do
+        visit "/t/#{topic.id}"
+
+        topic_page.click_assign_topic
+        assign_modal.set_assignee(staff_user)
+        assign_modal.confirm
+
+        expect(find("#post_2")).to have_content("Assigned")
+
+        find(".topic-footer-main-buttons .toggle-admin-menu").click
+        find(".topic-admin-close").click
+
+        expect(find("#post_3")).to have_content("Closed")
+        expect(page).not_to have_css("#post_4")
+        expect(page).not_to have_css("#topic .assigned-to")
+      end
+
+      it "can assign the previous assignee" do
+        visit "/t/#{topic.id}"
+
+        topic_page.click_assign_topic
+        assign_modal.set_assignee(staff_user)
+        assign_modal.confirm
+
+        expect(find("#post_2")).to have_content("Assigned")
+
+        find(".topic-footer-main-buttons .toggle-admin-menu").click
+        find(".topic-admin-close").click
+
+        expect(find("#post_3")).to have_content("Closed")
+        expect(page).not_to have_css("#post_4")
+        expect(page).not_to have_css("#topic .assigned-to")
+
+        topic_page.click_assign_topic
+        assign_modal.set_assignee(staff_user)
+        assign_modal.confirm
+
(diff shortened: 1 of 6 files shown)


jjaffeux:
@jancernik would you want to try to write a system spec for this? I can offer guidance and help if needed.
You would just have to write the first spec which would check the state when assigning in a closed topic.

jancernik:
Yeah sure, I'll take a look

jjaffeux:
> Yeah sure, I'll take a look

This is the commit where it's been introduced for encrypt, it's the most similar to what you have here:

https://github.com/discourse/discourse-encrypt/commit/b073a654e7812fdab74724e7a5d74416ded6320e

spec/system/assign_topic_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+
+  describe "with open topic" do
+    it "can assign and unassign" do
+      visit "/t/#{topic.id}"
+
+      topic_page.click_assign_topic
+      assign_modal.set_assignee(staff_user)
+      assign_modal.confirm
+
+      expect(find("#post_2")).to have_content("Assigned")
```
jjaffeux, line 24: Probably want to use I18n for all these strings

spec/system/assign_topic_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+      expect(find("#topic .assigned-to")).to have_content(staff_user.username)
+
+      topic_page.click_unassign_topic
+
+      expect(find("#post_3")).to have_content("Unassigned")
+      expect(page).not_to have_css("#topic .assigned-to")
```
jjaffeux, line 30: This one is dangerous. When you have this form it will never wait for it (2 seconds) to show before agreeing it's not on screen.

You should write it like this:

```
expect(page).to have_no_css("#topic .assigned-to")
```

This form will wait from 0 to 2s to declare it's not existing. If it's not existing from 0, then it will be satisfied right away.

spec/system/assign_topic_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+        assign_modal.confirm
+
+        expect(page).to have_no_css("#post_4")
+        expect(find("#topic .assigned-to")).to have_content(staff_user.username)
+      end
+
+      context "when reassign_on_open is set to true" do
+        before { SiteSetting.reassign_on_open = true }
+
+        it "topic gets reassigned on open" do
```
jjaffeux, line 84:
```suggestion
        it "reassigns the topic on open" do
```

spec/system/assign_topic_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+      topic_page.click_unassign_topic
+
+      expect(topic_page).to have_assignment_action(3, "unassigned", staff_user)
+      expect(page).to have_no_css("#topic .assigned-to")
+    end
+
+    context "when unassign_on_close is set to true" do
+      before { SiteSetting.unassign_on_close = true }
+
+      it "topic gets unassigned on close" do
```
jjaffeux, line 36:
```suggestion
      it "unassigns the topic on close" do
```

spec/system/assign_topic_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+        find(".topic-admin-close").click
+
+        expect(find("#post_3")).to have_content(
+          I18n.t("js.action_codes.closed.enabled", when: "just now"),
+        )
+        expect(page).to have_no_css("#post_4")
+        expect(page).to have_no_css("#topic .assigned-to")
+
+        topic_page.click_assign_topic
+        assign_modal.set_assignee(staff_user)
```
jjaffeux, line 74: Maybe we could use ruby at our advantage here and defined `def assignee=(user)` instead of `set_assignee`

spec/system/assign_topic_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+
+  describe "with open topic" do
+    it "can assign and unassign" do
+      visit "/t/#{topic.id}"
+
+      topic_page.click_assign_topic
+      assign_modal.set_assignee(staff_user)
+      assign_modal.confirm
+
+      expect(topic_page).to have_assignment_action(2, "assigned", staff_user)
```
jjaffeux, line 24: We could probably come up with more explicit functions here. Maybe:

```
have_assigned(staff_user) # would default to current topic?
have_assigned(staff_user, at_post: 2)
have_unassigned(staff_user, at_post: 3)
```

jjaffeux:
This is really good work, just added minor comments :100:

jjaffeux (approved):

spec/system/page_objects/pages/topic.rb
```diff
@@ -17,10 +17,19 @@ def click_edit_topic_assignment
         find("[data-value='reassign']").click
       end
 
-      def has_assignment_action?(post_num, action, assignee)
-        assignee = assignee.is_a?(Group) ? assignee.name : assignee.username
-        find("#post_#{post_num}").has_content?(
-          I18n.t("js.action_codes.#{action}", who: "@#{assignee}", when: "just now"),
+      def has_assigned?(args)
```
jjaffeux, line 20: :chef_kiss:

jancernik:
Thanks! Any more thoughts?

jjaffeux:
@jancernik the only missing thing is to create the workflow for github, at the moment the system tests are not run on the CI, look at what I linked to you in encrypt at the beginning.

jancernik:
We need to update [discourse-plugin.yml](https://github.com/discourse/.github/blob/main/.github/workflows/discourse-plugin.yml) to include system tests since now we have [reusable-workflows](https://github.com/discourse/discourse-assign/commit/4eeee568afa7279645494a27b957756565253460/)

jjaffeux:
FWIW Im fine if you merge it as it is. Would be nice to follow up with trying to have system tests working though.

pmusaraj:
Yeah, agreed, adding system tests to CI shouldn't block merging this PR. LGTM!

PR #427, opened 2023-01-12
FEATURE: Automatically open assignee chooser

nbianca:
Automatically expand the assignee user chooser when showing the assign modal.


jjaffeux:
This will only work for newest discourse instance, and do nothing for older installs. I guess you know this, but just noting it.

jjaffeux (approved):

PR #428, opened 2023-01-13
FIX: Deactivate active assignments attached to deleted targets

s3lase:
Before the implementation of the `post_destroyed` event handler in the plugin, soft deleting assigned posts/topics did not deactivate the assignment.
This left sites which assigned and deleted posts/topics before the introduction of the event handler with 'orphaned' assignments in the UI.

This change deactivates these orphaned assignments.

### DIFF
--- a/db/migrate/20230113025043_deactivate_assignments_to_deleted_targets.rb
+++ b/db/migrate/20230113025043_deactivate_assignments_to_deleted_targets.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+class DeactivateAssignmentsToDeletedTargets < ActiveRecord::Migration[7.0]
+  def up
+    execute <<~SQL
+      UPDATE assignments
+      SET active = false
+      FROM posts
+      WHERE posts.id = assignments.target_id
+        AND assignments.target_type = 'Post'
+        AND posts.deleted_at IS NOT NULL
+        AND assignments.active = true
+    SQL
+
+    execute <<~SQL
+      UPDATE assignments
+      SET active = false
+      FROM topics
+      WHERE topics.id = assignments.target_id
+        AND assignments.target_type = 'Topic'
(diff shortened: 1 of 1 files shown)


db/migrate/20230113025043_deactivate_assignments_to_deleted_targets.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+class DeactivateAssignmentsToDeletedTargets < ActiveRecord::Migration[7.0]
+  def up
+    execute <<~SQL
+      UPDATE assignments
+      SET active = false
+      FROM posts
+      WHERE posts.id = assignments.target_id
+        AND assignments.target_type = 'Post'
```
ZogStriP, line 10: Do we need this check? What would happen if we were to remove it? That would combine both queries into one, but does it has other side effects? 🤔

s3lase: The `assignments.target_type = 'Post'` condition isn't necessary if we're sticking with  the separate queries. 
However, removing it won't allow for the queries to be combined, we'll still need these 'join' conditions in place to  select the right assignments for update.

On the subject of combining these, I considered using CTE or subquery approach but decided to go with this in the interest of simplicity and readability.

I've update the PR to combine these via explicit joins. Do take a look and let know what you think.

ZogStriP: Oh right, I'm a dumbass 🤦 

Sorry, but I think the two queries were much simpler and easier to read than the one query with two `LEFT JOIN`s.

Up to you if you want to keep it or revert back to two queries.

s3lase: Yes. I agree. The two queries are more readable

PR #429, opened 2023-01-13
UX: Add suggestions label to assign modal

nbianca:
New design:

image

### DIFF
--- a/assets/javascripts/discourse/templates/components/assignee-chooser-row.hbs
+++ b/assets/javascripts/discourse/templates/components/assignee-chooser-row.hbs
@@ -1,0 +1,19 @@
+{{#if this.item.isUser}}
+  {{avatar this.item imageSize="tiny"}}
+  <span class="identifier">{{format-username this.item.id}}</span>
+  <span class="name">{{this.item.name}}</span>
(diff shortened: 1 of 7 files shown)


jjaffeux:
@nbianca linting error is legit

assets/javascripts/discourse/templates/components/assignee-chooser-row.hbs
```diff
@@ -0,0 +1,19 @@
...
+  {{decorate-username-selector this.item.id}}
+{{else if this.item.isGroup}}
+  {{d-icon "users"}}
+  <span class="identifier">{{this.item.id}}</span>
+  <span class="name">{{this.item.full_name}}</span>
+{{else}}
+  {{d-icon "envelope"}}
+  <span class="identifier">{{this.item.id}}</span>
+{{/if}}
+{{#if this.item.isSuggestion}}
```
jjaffeux, line 17: I assume this is not yet using the new sk row index right?

nbianca: Unfortunately no because the text has to be displayed for the first element and the first search without a term.

pmusaraj (approved):

PR #430, opened 2023-01-17
FIX: System tests failing due to assignee chooser

jancernik:
Since now the assignee chooser opens automatically, clicking on it was causing it to close


jjaffeux (approved):
Good catch :+1:

PR #441, opened 2023-02-10
UX: improve custom icons for group assign

awesomerobot:
I've made some new icons that I think work better alongside the existing assign icons. 

The previous group assign icon wasn't as tall as the single assign icon, so it could look slightly misaligned on topic lists. The new group assign icons share the same baseline, height, and keeps the +/x at the same size and position as the single assign. Much more harmonious ☯️ 

At large sizes the old group assign icon feels a little too large and awkward (dissing my own design here 💅🏻). The new ones fit much better. 

Before:

Screenshot 2023-02-10 at 4 18 46 PM

After:

Screenshot 2023-02-10 at 4 19 32 PM

Before: 

Screenshot 2023-02-10 at 4 18 33 PM

After:

Screenshot 2023-02-10 at 4 18 00 PM


pmusaraj (approved):
Nice!

PR #446, opened 2023-02-14
DEV: Update plugin to match latest guidelines

nbianca:
- Define extension modules
- Use different files instead of plugin.rb
- Make sure plugin is disabled according to the setting


app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -279,17 +271,16 @@ def recent_assignees
             HAVING COUNT(*) < #{SiteSetting.max_assigned_topics}
           ) as X ON X.user_id = users.id
         SQL
-        .joins(<<~SQL)
+      .joins(<<~SQL)
           LEFT JOIN(
             SELECT DISTINCT ON (user_id) name, user_id
             FROM user_custom_fields
-            WHERE name = '#{DiscourseCalendar::HOLIDAY_CUSTOM_FIELD}'
+            WHERE name = '#{DiscourseAssign::DiscourseCalendar::HOLIDAY_CUSTOM_FIELD}'
```
CvX, line 278: I was like: this isn't correct, is it? 😄 (it is 😄)

lib/discourse_assign/group_extension.rb
```diff
@@ -0,0 +1,23 @@
...
+                "assignable_level in (:levels) OR
+                  (
+                    assignable_level = #{Group::ALIAS_LEVELS[:members_mods_and_admins]} AND id in (
+                    SELECT group_id FROM group_users WHERE user_id = :user_id)
+                  ) OR (
+                    assignable_level = #{Group::ALIAS_LEVELS[:owners_mods_and_admins]} AND id in (
+                    SELECT group_id FROM group_users WHERE user_id = :user_id AND owner IS TRUE)
+                  )",
+                levels: alias_levels(user),
+                user_id: user && user.id,
```
CvX, line 18:
```suggestion
                user_id: user&.id,
```

CvX (approved):

PR #452, opened 2023-03-10
UX: clearer assign messages on first post

awesomerobot:
This is a little more verbose, but fixes the grammar for assigning to post and specifies topic assigns

`Assigned to #2 discobot` -> `Assigned #2 to discobot`

`Assigned to awesomerobot, #2 discobot` -> `Assigned topic to awesomerobot, #2 to discobot`

Before:
Screenshot 2023-03-10 at 4 43 59 PM

After: 
Screenshot 2023-03-10 at 4 40 37 PM

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -27,6 +27,8 @@ en:
       assigned: "Assigned"
       group_everyone: "Everyone"
       assigned_to: "Assigned to"
+      topic_to: "topic to"
+      assign_post_to: "#%{post_number} to %{username}"
       assigned_to_w_ellipsis: "Assigned to..."
       assign_notification: "<p><span>%{username}</span> %{description}</p>"
       assign_group_notification: "<p><span>%{username}</span> %{description}</p>"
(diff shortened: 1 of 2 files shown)


config/locales/client.en.yml
```diff
@@ -27,6 +27,8 @@ en:
       assigned: "Assigned"
       group_everyone: "Everyone"
       assigned_to: "Assigned to"
+      topic_to: "topic to"
```
CvX, line 30: Paging doctor @gschlager - as I understand, partial strings like that aren't nicely translatable and there should be one (or more) complete string that's a full message?

ZogStriP: I agree, this one might be tricky enough to require the more advanced Message Format string?

gschlager: Hmm, this is hard to solve correctly unless widgets allow something like `htmlSafe` so that we can put the whole string into the locale file. Correct me if I'm wrong, but I don't think we have something like that for widgets.

Aren't we planning to replace widgets in the near future? I guess in that case it's fine to accept this PR as is. But I'd recommend adding comments to the new strings in the locale file so that translators know what comes before and after those strings in the UI.

awesomerobot: Thanks! I've added a couple comments to help clarify

awesomerobot: Actually, on second thought I tried another way with more complete strings

CvX (approved):
👍, but still would be nice to make it work for RTL langs.
Maybe we'll tackle that when we'll be rewriting all of this widget code 🙂

PR #457, opened 2023-03-30
FIX: Show user status tooltip on the assign modal

jancernik:
This PR needs https://github.com/discourse/discourse/pull/20897

UserStatusMessage


AndrewPrigorshnev:
We may not need this PR if we'll be able to address this comment in another PR https://github.com/discourse/discourse/pull/20897#discussion_r1160841448, we may stop event propagation in the `user-status-message` component or even better in the `d-tooltip` component.

AndrewPrigorshnev (approved):

PR #461, opened 2023-04-27
DEV: Clean up `remind-assigns-frequency` logic

janzenisaac:


assets/javascripts/discourse/templates/components/remind-assigns-frequency.hbs
```diff
@@ -6,7 +6,7 @@
       @valueProperty="value"
       @content={{availableFrequencies}}
       @value={{selectedFrequency}}
-      @onChange={{action (mut user.custom_fields.remind_assigns_frequency)}}
+      @onChange={{action "updateSelectedFrequency"}}
```
CvX, line 9: We're gonna codemod away the `action` helper at some point, but if we're already changing this line…

```suggestion
      @onChange={{this.updateSelectedFrequency}}
```

assets/javascripts/discourse/components/remind-assigns-frequency.js
```diff
@@ -1,26 +1,24 @@
...
 export default Component.extend({
-  @discourseComputed(
-    "user.custom_fields.remind_assigns_frequency",
-    "siteSettings.remind_assigns_frequency"
-  )
-  selectedFrequency(userAssignsFrequency, siteDefaultAssignsFrequency) {
+  siteSettings: service(),
+
+  get selectedFrequency() {
+    const frequency = this.user.get("custom_fields.remind_assigns_frequency");
```
CvX, line 10: I'm pretty sure this (and the other getter) has to be a `@computed` property still as it relies on a non-autotracked property on a user model.

PR #463, opened 2023-04-28
FIX: properly check key values for shortcuts

janzenisaac:
- Update angle bracket components to use vanilla attributes where possible
- Update how we check for key value after changing to `on "change"` from `onChange`

### DIFF
--- a/assets/javascripts/discourse-assign/connectors/groups-interaction-custom-options/assignable-interaction-fields.hbs
+++ b/assets/javascripts/discourse-assign/connectors/groups-interaction-custom-options/assignable-interaction-fields.hbs
@@ -11,7 +11,7 @@
     @valueProperty="value"
     @value={{assignableLevel}}
     @content={{assignableLevelOptions}}
-    class="groups-form-assignable-level"
+    @class="groups-form-assignable-level"
     @onChange={{action (mut model.assignable_level)}}
   />
 </div>
--- a/assets/javascripts/discourse-assign/connectors/groups-interaction-custom-options/assignable-interaction-fields.hbs
+++ b/assets/javascripts/discourse-assign/connectors/groups-interaction-custom-options/assignable-interaction-fields.hbs
@@ -11,7 +11,7 @@
     @valueProperty="value"
     @value={{assignableLevel}}
     @content={{assignableLevelOptions}}
-    @class="groups-form-assignable-level"
+    class="groups-form-assignable-level"
     @onChange={{action (mut model.assignable_level)}}
   />
 </div>
--- a/assets/javascripts/discourse-assign/controllers/assign-user.js
+++ b/assets/javascripts/discourse-assign/controllers/assign-user.js
@@ -72,8 +72,8 @@ export default Controller.extend(ModalFunctionality, {
   },
 
   @action
-  handleTextAreaKeydown(value) {
-    if ((value.ctrlKey || value.metaKey) && value.key === "Enter") {
+  handleTextAreaKeydown(event) {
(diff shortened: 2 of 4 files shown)


assets/javascripts/discourse-assign/connectors/groups-interaction-custom-options/assignable-interaction-fields.hbs
```diff
@@ -11,7 +11,7 @@
     @valueProperty="value"
     @value={{assignableLevel}}
     @content={{assignableLevelOptions}}
-    @class="groups-form-assignable-level"
+    class="groups-form-assignable-level"
```
CvX, line 14: I don't this works. d-button has explicitly added `...attributes` to its element, but select-kit's combo box doesn't have that

davidtaylorhq: For classic components, attributes are added to the wrapper element (assuming `tagName` is not overridden to `""`)

assets/javascripts/discourse/templates/modal/assign-user.hbs
```diff
@@ -60,16 +60,16 @@
 </DModalBody>
 
 <div class="modal-footer">
-  {{d-button
-    label=(if
+  <DButton
+    @label={{(if
```
jjaffeux, line 64: Shouldnt this just be `{{if ...`

assets/javascripts/discourse/templates/modal/assign-user.hbs
```diff
@@ -60,16 +60,16 @@
...
       model.reassign
       "discourse_assign.reassign.title"
       "discourse_assign.assign_modal.assign"
-    )
-    icon=inviteIcon
-    action=(action "assign")
+    )}}
+    @icon={{inviteIcon}}
+    @action={{(action "assign")}}
```
jjaffeux, line 70: Same question about `{{}}` and also is it possible to do `this.assign` here instead or it doesnt work in this case?

assets/javascripts/discourse/templates/components/remind-assigns-frequency.hbs
```diff
@@ -2,7 +2,7 @@
   <div class="controls controls-dropdown">
     <label>{{i18n "discourse_assign.reminders_frequency.description"}}</label>
     <ComboBox
-      @id="remind-assigns-frequency"
+      id="remind-assigns-frequency"
```
CvX, line 5: the lack of `...attributes` applies to all attrs (yeah, select-kit is not glimmer ready 😉)

assets/javascripts/discourse-assign/controllers/assign-user.js
```diff
@@ -72,8 +72,8 @@ export default Controller.extend(ModalFunctionality, {
   },
 
   @action
-  handleTextAreaKeydown(value, event) {
-    if ((event.ctrlKey || event.metaKey) && event.key === "Enter") {
+  handleTextAreaKeydown(value) {
```
jjaffeux, line 75: So value is actually an event here right?

assets/javascripts/discourse-assign/controllers/assign-user.js
```diff
@@ -72,8 +72,8 @@ export default Controller.extend(ModalFunctionality, {
...
+    if ((value.ctrlKey || value.metaKey) && value.key === "Enter") {
```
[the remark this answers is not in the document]
jjaffeux: just made the same comment 😅

CvX (approved):

PR #464, opened 2023-05-01
DEV: Remove old user menu related code

keegangeorge:
In line with the removal of the legacy user menu (https://github.com/discourse/discourse/pull/21308), this PR updates this plugin to also remove anything related to the old user menu.


test/javascripts/acceptance/assigns-tab-user-menu-test.js
```diff
@@ -215,9 +215,9 @@ acceptance(
 );
 
 acceptance(
-  "Discourse Assign | experimental user menu | assign_enabled setting is disabled",
+  "Discourse Assign | user menu | assign_enabled setting is disabled",
   function (needs) {
-    needs.user({ redesigned_user_menu_enabled: true, can_assign: true });
```
davidtaylorhq, line 220: If we restore the `redesigned_user_menu_enabled; true`, will that allow us to merge this PR before merging the core change?

assets/javascripts/discourse-assign/initializers/extend-for-assigns.js
```diff
@@ -658,24 +657,6 @@ function initialize(api) {
     }
   });
 
-  api.addUserMenuGlyph((widget) => {
```
davidtaylorhq, line 661: We should add a `.discourse-compatibility` entry so that an older version of the plugin is used on Discourse 3.1.0.beta3 or below.

davidtaylorhq:
Can we delete the files under [discourse/widgets](https://github.com/discourse/discourse-assign/tree/main/assets/javascripts/discourse/widgets) ?

assets/javascripts/discourse-assign/initializers/extend-for-assigns.js
keegangeorge: There seems to be one already present from this recent commit: https://github.com/discourse/discourse-assign/pull/458

PR #465, opened 2023-05-01
DEV: Remove usage of `tagName`

janzenisaac:


davidtaylorhq:
What's the context of this? Was it a deprecation? 🤔

davidtaylorhq:
Yeah... `tagName` here is consumed by the raw-hbs template:

https://github.com/discourse/discourse/blob/a1b35601fcd209d83471e269e23b7c4f9a42465a/app/assets/javascripts/discourse/app/templates/list/activity-column.hbr#L1-L1

So I think failing to pass it will lead to some weird/broken behaviour.

PR #467, opened 2023-05-04
DEV: Replace #pluck_first monkey patch with native #pick

Drenmi:
### What is this change?

We have replaced the `#pluck_first` freedom patch with the native `#pick` in core [here](https://github.com/discourse/discourse/pull/19893).

This change is updates this plugin in the same fashion.


gschlager:
You might need to update the `.discourse-compatibility` file if this change is incompatible with older Discourse versions.

tgxworld (approved):

PR #468, opened 2023-05-04
FIX: Reassigning to a group creates an incorrect mention's link

Arkshine:
_Reported here_: https://meta.discourse.org/t/reassigning-to-a-group-creates-a-u-link-in-the-group-mention/263659

Reassigning to a group creates an `/u/` link in the `@group` mention instead of `/g/`.

image

The reason is the `reassigned_group` group small post action code is not registered, thus creating a user link by default.

Related Discourse code:
https://github.com/discourse/discourse/blob/main/app/assets/javascripts/discourse/app/widgets/post-small-action.js#L12-L27
https://github.com/discourse/discourse/blob/main/app/assets/javascripts/discourse/app/widgets/post-small-action.js#L82-L84

### DIFF
--- a/assets/javascripts/discourse-assign/initializers/extend-for-assigns.js
+++ b/assets/javascripts/discourse-assign/initializers/extend-for-assigns.js
@@ -974,6 +974,7 @@ export default {
 
     withPluginApi("0.12.2", (api) => {
       api.addGroupPostSmallActionCode("assigned_group");
+      api.addGroupPostSmallActionCode("reassigned_group");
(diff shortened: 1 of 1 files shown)


SamSaffron (approved):

SamSaffron:
generally we prefer tests for all changes, I am giving this a pass though cause it feels like there is a huge subset there that is missing tests and putting this all on you is unfair.

Arkshine:
Oh. I was wondering If It was necessary to add tests in this case. I have my answer! I'm not used to making tests; I appreciate your sparing me. 😄

PR #469, opened 2023-05-17
DEV: Don't use renderTemplate

CvX:
See: https://github.com/discourse/discourse/pull/21517


test/javascripts/acceptance/user-assignments-list-test.js
CvX, line 1: that test wasn't testing 🤪

.github/workflows/discourse-plugin.yml
```diff
@@ -9,3 +9,5 @@ on:
 jobs:
   ci:
     uses: discourse/.github/.github/workflows/discourse-plugin.yml@v1
+    with:
+      core_ref: a-template-name
```
davidtaylorhq, line 13: Don't forget to remove before merging :P

davidtaylorhq (approved):

PR #474, opened 2023-05-26
FEATURE: skip group invite if all members can see topic already.

vinothkannans:
We don't want to invite a group when all of its members are part of another group that is already allowed to the topic.

### DIFF
--- a/lib/assigner.rb
+++ b/lib/assigner.rb
@@ -461,10 +461,13 @@ class ::Assigner
 
   def invite_group(group)
     return if topic.topic_allowed_groups.exists?(group_id: group.id)
-    return if topic.all_allowed_users
-                .joins("RIGHT JOIN group_users ON group_users.user_id = users.id")
-                .where("group_users.group_id = ? AND users.id IS NULL", group.id)
-                .empty? # all group members can already see the topic
+    if topic
+         .all_allowed_users
+         .joins("RIGHT JOIN group_users ON group_users.user_id = users.id")
+         .where("group_users.group_id = ? AND users.id IS NULL", group.id)
+         .empty?
+      return # all group members can already see the topic
+    end
 
     guardian.ensure_can_invite_group_to_private_message!(group, topic)
     topic.invite_group(@assigned_by, group)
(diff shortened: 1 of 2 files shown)


lib/assigner.rb
```diff
@@ -463,6 +452,31 @@ def unassign(silent: false, deactivate: false)
...
+    topic.invite(@assigned_by, user.username)
+  end
+
+  def invite_group(group)
+    return if topic.topic_allowed_groups.exists?(group_id: group.id)
+    if topic
+         .all_allowed_users
+         .joins("RIGHT JOIN group_users ON group_users.user_id = users.id")
+         .where("group_users.group_id = ? AND users.id IS NULL", group.id)
+         .empty?
```
tgxworld, line 468:
```suggestion
         .exists?
```

Small suggestion but I think we can be more efficient here by just doing a `SELECT 1` query

spec/lib/assigner_spec.rb
```diff
@@ -740,17 +740,41 @@ def assigned_to?(assignee)
...
+        3.times.collect do
+          Fabricate(
+            :group,
+            assignable_level: Group::ALIAS_LEVELS[:only_admins],
+            messageable_level: Group::ALIAS_LEVELS[:only_admins],
+          )
+        end
+      group1.add(user1)
+      group1.add(user3)
+      group2.add(user2)
```
tgxworld, line 760:
```suggestion
      group2.add(user2, user3)
```
Small suggestion to improve the setup of the test here.

tgxworld (approved):
Some small suggestions but changes look good to me.

lib/assigner.rb
vinothkannans: I'm still using `SELECT 1` here. The [`empty?`](https://apidock.com/rails/v5.2.3/ActiveRecord/Relation/empty%3F) method is equal to the `!exists?` method.

tgxworld: O icic. I thought `empty?` is an array method. All good here :+1: Thank you for clarifying.

PR #477, opened 2023-06-01
UX: constrain avatar size due to core change

awesomerobot:
We reduced the number of avatar sizes in https://github.com/discourse/discourse/commit/c2332d7505379c30e11d295d90f5224385736993, so the avatar in the assign footer button was too large and causing the button to be too tall. This centers it and shrinks the avatar back to the previous size (20px square)

### DIFF
--- a/assets/stylesheets/assigns.scss
+++ b/assets/stylesheets/assigns.scss
@@ -176,10 +176,14 @@
 
 #topic-footer-dropdown-reassign {
   .name {
-    font-weight: normal !important;
+    font-weight: normal;
+    display: flex;
+    align-items: center;
   }
   .avatar {
     margin-right: 0.25em;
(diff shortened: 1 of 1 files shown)


jjaffeux (approved):

assets/stylesheets/assigns.scss
```diff
@@ -176,10 +176,14 @@
 
 #topic-footer-dropdown-reassign {
   .name {
-    font-weight: normal !important;
+    font-weight: normal;
+    display: flex;
```
CvX, line 180: @awesomerobot this regressed the spacing on `"Unassign… @group"` label:

before/after

Screenshot 2023-07-06 at 22 48 47 Screenshot 2023-07-06 at 22 51 58

PR #482, opened 2023-06-29
FIX: Sum assignments for group and group users

nattsw:
Related: https://meta.discourse.org/t/bad-group-assignment-count/252636

Use the `Assignment` count rather than the number of `Topic` assigned. 

e.g. John can be assigned to topic_id: 1, and Jane can be assigned to a post within topic_id: 1. This will result in 1 rather than 2.

Before:

The sum doesn't add up.

image

There's a flicker in the number in the nav.

https://github.com/discourse/discourse-assign/assets/1555215/786d18ee-1505-41ff-b543-2cd81ce2ff09

After:

The nav number will just be a sum of all assignments, user+group.

https://github.com/discourse/discourse-assign/assets/1555215/12f28b66-238b-4445-8229-127e73825e52

There is still another bug we are seeing where it's likely that this line with the `limit` is causing issues with accuracy. But that will be looked at in another PR.

https://github.com/discourse/discourse-assign/blob/2bff29e919bc21d610c24c75299bd97090052d04/app/controllers/discourse_assign/assign_controller.rb#L128-L141

### DIFF
--- a/app/controllers/discourse_assign/assign_controller.rb
+++ b/app/controllers/discourse_assign/assign_controller.rb
@@ -125,7 +125,7 @@ module DiscourseAssign
 
       guardian.ensure_can_see_group_members!(group)
 
-      members =
+      users_with_assignments_count =
         User
           .joins("LEFT OUTER JOIN group_users g ON g.user_id = users.id")
           .joins(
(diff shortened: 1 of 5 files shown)


lis2 (approved):

nattsw:
Thank you @lis2

PR #484, opened 2023-07-03
DEV: Merge js dirs; prefix tests

CvX:


nattsw (approved):

nattsw:
Btw, do we have a `.git-blame-ignore-revs` for this repo?

CvX:
GH’s UI handles renames well now so it might not be necessary. 😃

(I’ve been burned by ignore-revs 😅 no commit is ever irrelevant so having some of them disappear sometimes bites you back 😛 and the way we once introduced a small syntax error in ignore-revs in core long time ago? 😄 now you end up on an error page if you blame-surf somewhere between the broken commit and the fix 🥲)

PR #487, opened 2023-07-05
DEV: Fabricate assignments

nattsw:
Hopefully makes things less flakey by
- Not fabricating in `before`

Also
- Use fabricators instead of the `Assigner` to set up tests
- Be specific about users and groups


nattsw:
Thank you @ZogStriP

PR #491, opened 2023-07-11
DEV: Use the new bulk topic actions api

CvX:


CvX:
I re-added the old code paths so it can be merged before the core PR (and the build can remain green throughout the upgrade)

davidtaylorhq (approved):

davidtaylorhq (approved):

PR #492, opened 2023-07-12
DEV: Try to make system tests more stable

tgxworld:
Why this change?

We have been seeing tests failures due to errors like

```
Capybara::ElementNotFound:
  Unable to find css "#topic-footer-button-assign"
```

when running our system tests in CI. However, I can't quite figure out
why that is happening and have invested way too much time to do so.
Therefore, I'm trying to fix this by closing a topic via the topic
timeline controls instead of the footer button instead.

Example of test failure: https://github.com/discourse/discourse/actions/runs/5525940028/jobs/10080088908


martin-brennan (approved):
Such a weird one

tgxworld:
Thank you for the review @martin-brennan

PR #496, opened 2023-08-08
FIX: Fix reminder frequency not getting sent daily when early by a few seconds/minutes

nattsw:
t/105676

When a user sets their reminder frequency to be "daily", there's a chance that they actually only get it sent every two days.

 (/my/preferences/notifications)
Screenshot 2023-08-08 at 7 50 15 PM

This is likely happening as our `EnqueueReminders` sidekiq job runs every `1.day`, issuing a second job `RemindUser` which updates the `last_reminded_at` value to be a bit after the time when the sidekiq job gets done. The list of users to send reminders to depends on `last_reminded_at` being more than the user set reminder frequency, resulting in the bug that the reminder only triggers on the next day.

### DIFF
--- a/spec/jobs/scheduled/enqueue_reminders_spec.rb
+++ b/spec/jobs/scheduled/enqueue_reminders_spec.rb
@@ -3,12 +3,13 @@
 require "rails_helper"
 
 RSpec.describe Jobs::EnqueueReminders do
-  let(:assign_allowed_group) { Group.find_by(name: "staff") }
-  let(:user) { Fabricate(:user, groups: [assign_allowed_group]) }
+  fab!(:assign_allowed_group) { Fabricate(:group) }
+  fab!(:user) { Fabricate(:user, groups: [assign_allowed_group]) }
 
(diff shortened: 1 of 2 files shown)


spec/jobs/scheduled/enqueue_reminders_spec.rb
```diff
@@ -36,55 +37,69 @@
...
 
-      assert_reminders_enqueued(1)
-    end
+        assign_multiple_tasks_to(user, assigned_on: 1.day.ago - 1.minute)
 
-    it "does not enqueue reminders if the remind frequency is set to never" do
-      SiteSetting.remind_assigns_frequency = 0
-      assign_multiple_tasks_to(user)
+        assert_reminders_enqueued(1)
+      end
```
nattsw, line 61: Non-obvious, but this is the new test for the buffer.

Drenmi (approved):
🕙

nattsw:
Thank you @Drenmi

PR #499, opened 2023-08-15
DEV: Increase the buffer used to determine whether to send out the reminder email

nattsw:
Continuing from https://github.com/discourse/discourse-assign/pull/496

The buffer value introduced here was not enough. We are seeing that the deviation across emails across the two days (when it should be one day) is two hours. 

Screenshot 2023-08-15 at 1 39 52 PM

Day 1 email at 5:47PM. Assuming the next sidekiq job triggers at 5:46PM (at an earlier time the next day), the reminder will not be sent

Screenshot 2023-08-15 at 1 39 48 PM

Day 3 email at 4:19PM, about 2 hours earlier in the day. 

So we will use 2 hours as a buffer.

### DIFF
--- a/app/jobs/scheduled/enqueue_reminders.rb
+++ b/app/jobs/scheduled/enqueue_reminders.rb
@@ -20,7 +20,7 @@ module Jobs
       Group.assign_allowed_groups.pluck(:id).join(",")
     end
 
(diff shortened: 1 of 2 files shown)


lis2 (approved):

nattsw:
Thank you @lis2

PR #507, opened 2023-09-07
FIX: Display assignments in user menu properly

Flink:
Currently, we display a mix of topics and notifications in the user menu assignments tab. This has a number of issues like having to maintain hard to understand code instead of simply relying on the notifications system we have, we can’t display more than one assignment per topic and it’s not clear if an assignment is for a topic or a post nor if it’s a group assignment or an individual one.
    
This PR addresses those issues by relying on the notifications system we’re using for most of the other user menu tabs instead of a custom implementation. This led to some heavy refactoring but it was worthwhile as things are a bit more normalized and easier to reason about. Instead of serializing topics with different attributes to access various assignments, we now simply return a notification for each existing assignment.
    
The UI changed a bit: tooltips are now explaining if the assignment is for a topic or a post and if it’s for an individual or a group. Icons are also properly set (so no more individual icon for group assignments) and the assigned group name is displayed.
    
The background jobs signatures changed a bit (`assignment_id` is now needed for the unassign job) so when deploying this patch, it’s expected to have some jobs failing. That’s why a post-migration has been written to handle the creation of missing notifications and the deletion of extra notifications too.


rishabhnambiar:
I'll wave to @nlalonde / @heymarkreeves for a review when they have space 👋

nlalonde (approved):

PR #509, opened 2023-09-12
FIX: Don't crash on action posts without a label

Firepup6500:
Adds a check to prevent assign from trying to act on action posts missing a label. Likely not a common issue, but better safe than a crash.


Firepup6500:
[This topic](https://meta.discourse.org/t/weird-bug-with-discourse-assign/276109) preempted this PR.

davidtaylorhq:
Plugin moving to core

PR #522, opened 2023-11-09
FIX: Order items properly in user menu

Flink:
The recent changes to the user menu changed how we were ordering the items in it. A small bug was noticed though, as we’re not displaying the unread assignments at the very top.

This PR fixes the ordering like this:

- Unread individual assignment notifications
- Unread group assignment notifications
- Read individual assignments
- Read group assignments

In each group of items, they are sorted by date, the most recent ones being at the top.

### DIFF
--- a/assets/javascripts/discourse/components/user-menu/assigns-list.js
+++ b/assets/javascripts/discourse/components/user-menu/assigns-list.js
@@ -1,3 +1,5 @@
+import { set } from "@ember/object";
+import { sort } from "@ember/object/computed";
 import UserMenuNotificationsList from "discourse/components/user-menu/notifications-list";
 import I18n from "I18n";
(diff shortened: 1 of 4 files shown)


mcwumbly:
if I understand correctly, we are only sorting on the front end. could that be an issue if there a many assignments? (not sure if we are guaranteed to have them all available to sort, or if we fetch them in smaller pages initially)

Flink:
No, we’re not fetching all the existing assignments, we’re relying on the notifications system. It means we’re fetching the 30 most recent assignments and by default the backend sorts them by unread first and then creation date. But since we want to group them, we have to do it on the frontend. I think having 30 assignments is already a lot, but you tell me :sweat_smile:

mcwumbly:
cool, sounds like things should mostly just work!

nlalonde (approved):

PR #542, opened 2024-01-04
DEV: Capture logs for all tests in `spec/system/assign_topic_spec.rb`

tgxworld:
Why this change?

We have been getting flaky test failures from these specs and the failure screenshot shows that the user is not logged in when it is supposed to be. Futher investigation shows that when the test flakes, it is because the request to view the topic is using an auth token that is different from the one which was created when the user was signed in.

What does this change do?

1. Add the `capture_log` metadata to all the tests in this file.

2. Enables the `verbose_auth_token_logging` site setting to give us more debugging information in the logs.

### DIFF
--- a/spec/system/assign_topic_spec.rb
+++ b/spec/system/assign_topic_spec.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-describe "Assign | Assigning topics", type: :system, js: true do
+describe "Assign | Assigning topics", type: :system, capture_log: true do
   let(:topic_page) { PageObjects::Pages::Topic.new }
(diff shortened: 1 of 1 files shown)


spec/system/assign_topic_spec.rb
```diff
@@ -10,6 +10,10 @@
 
   before do
     SiteSetting.assign_enabled = true
+
+    # The system tests in this file are flaky and auth token related so turning this on
+    SiteSetting.verbose_auth_token_logging = true
```
nattsw, line 15: 😿 

Lol would be funny if you turn it on and it stops flaking.

PR #546, opened 2024-02-07
DEV: Stop using FoundUserWithStatusSerializer

AndrewPrigorshnev:
Starting from https://github.com/discourse/discourse/commit/b3a1199493a5974353449e2b3f1b30d9eabbff59 we stop using deprecated "WithStatus" serializers. Instead we'll be passing an 'include_status' option to serilaizers, for example:

```ruby
# before
BasicUserWithStatusSerializer.new(user)
ArraySerializer.new(users, each_serializer: BasicUserWithStatusSerializer)

# now
BasicUserSerializer.new(user, include_status: true)
ArraySerializer.new(users, each_serializer: BasicUserSerializer, include_status: true)
```


jjaffeux:
We probably need a discourse compatibility entry too no?

AndrewPrigorshnev:
> We probably need a discourse compatibility entry too no?

Yeah, that's right. I was going to add it after we merge the Core PR.

nlalonde (approved):

PR #547, opened 2024-02-22
DEV:  Remove `capture_logs: true` metadata

tgxworld:
Why this change?

We believe we have figured out the problem and not longer need the debugging information.


Drenmi (approved):
💣

PR #556, opened 2024-03-26
FIX: Assignments limit shouldn't prevent from reassigning a post

AndrewPrigorshnev:
At the moment, it's possible to have maximum [5 assignments](https://github.com/discourse/discourse-assign/blob/08c2a4e0f392718fd821ce10a52186f04323e439/lib/assigner.rb#L7) per topic (that includes topic and post assignments). When trying to assign more, this message appears:

Screenshot 2024-03-26 at 21 09 49

One possible edge case here is reassigning a topic or a post. Reassignment doesn't lead to exceeding the limit, and therefore it should be possible. But at the moment we handle correctly only _topic_  reassignments, while when reassigning a _post_, we show the error message "the limit has been reached".

This PR makes post reassignments work correctly too.

### DIFF
--- a/spec/lib/assigner_spec.rb
+++ b/spec/lib/assigner_spec.rb
@@ -264,7 +264,7 @@ RSpec.describe Assigner do
         expect(second_assign[:success]).to eq(true)
       end
 
-      it "assignments limit doesn't prevent from reassigning a post" do
+      it "reassigns a post even when at the assignments limit" do
         posts =
           (described_class::ASSIGNMENTS_PER_TOPIC_LIMIT).times.map do
             Fabricate(:post, topic: topic)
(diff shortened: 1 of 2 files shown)


spec/lib/assigner_spec.rb
```diff
@@ -264,6 +264,21 @@ def assigned_to?(assignee)
         expect(second_assign[:success]).to eq(true)
       end
 
+      it "assignments limit doesn't prevent from reassigning a post" do
```
jjaffeux, line 267: thats not a very english sentence

ZogStriP, line 267: How about?

```suggestion
      it "reassigns even when at the assignments limit" do
```

ZogStriP (approved):

spec/lib/assigner_spec.rb
AndrewPrigorshnev: Sounds good, with one detail that we should explicitly mention that this test is for reassigning _a post_. We also reassign _topics_ sometimes.

PR #561, opened 2024-04-09
FIX: Unassign label sometimes touches avatar

AndrewPrigorshnev:
We've noticed that sometimes it looks like this:

f39e0c927da766956e46387ef5d10b73052ace3e_2_1380x338

Turned out that happens when a topic has several assignments and those contain assignments to users _and_ to groups. We don't show avatars for groups, but the code saw there are several assignees and applied overlapping styling.

I corrected the logic, and that fixed the issue.

Here is how avatars on this button behave. If there are from one to many assignments on the topic, but all the assignees are groups, we don't show avatars, only label:

Screenshot 2024-04-09 at 22 07 29

If there are from one to many assignments to the same user we show the avatar of that user (and even if there are also assignments to groups there is no overlapping):

Screenshot 2024-04-09 at 22 06 44

And if there are from two to many assignments to different users, we show avatars of first two users:

Screenshot 2024-04-09 at 22 07 11

### DIFF
--- a/assets/javascripts/discourse/components/topic-level-assign-menu.js
+++ b/assets/javascripts/discourse/components/topic-level-assign-menu.js
@@ -62,13 +62,7 @@ export default {
   },
 
   noneItem() {
-    const topic = this.topic;
-
-    if (topic.assigned_to_user || topic.hasAssignedPosts()) {
-      return unassignUsersButton(topic.uniqueAssignees());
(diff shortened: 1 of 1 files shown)


jjaffeux (approved):

pmusaraj (approved):
Very nice, thanks!

PR #562, opened 2024-04-12
FEATURE: Make it possible to reassign posts on the topic level assign modal

AndrewPrigorshnev:
This adds a new dropdown to the topic level assign modal. A topic may contain several assignments, the topic itself may be assigned and also some of the replies may be assigned. With this new dropdown, it's possible now to edit all the assignments in one go:

Screenshot 2024-04-12 at 20 39 52

Screenshot 2024-04-12 at 20 39 58


assets/javascripts/discourse/components/modal/edit-topic-assignments.js
```diff
@@ -0,0 +1,46 @@
...
+    try {
+      await this.#assign();
+    } catch (error) {
+      popupAjaxError(error);
+    }
+  }
+
+  async #assign() {
+    for (const assignment of this.assignments) {
+      if (assignment.isEdited) {
```
AndrewPrigorshnev, line 41: We don't send server requests for assignments that have not been edited. In case user edits several assignments, it will send several requests to the server. I'm not sure if it's worth introducing a server method here to do all that in one request, but it may be done in a separate PR in the future if case we decide we want it.

assets/javascripts/discourse/models/assignment.js
```diff
@@ -0,0 +1,38 @@
+import { tracked } from "@glimmer/tracking";
+import EmberObject from "@ember/object";
+
+export class Assignment extends EmberObject {
```
AndrewPrigorshnev, line 4: I've introduced this model on the client, but I haven't migrated all code to using it because that would extend the refactoring surface too much. It would be great to slowly migrate all code to using it, I think that will help simplify the code of the plugin. Now it contains only fields, but as soon as more code start using it, it will become possible to move some domain logic in here.

jjaffeux (approved):

PR #566, opened 2024-04-24
FIX: ensures ctrl+enter from textarea saves

jjaffeux:


jancernik (approved):

assets/javascripts/discourse/components/topic-assignments.gjs
```diff
@@ -41,7 +41,7 @@ export default class TopicAssignments extends Component {
     </div>
     <Assignment
       @assignment={{this.selectedAssignment}}
-      @onSubmit={{this.submit}}
+      @onSubmit={{this.args.onSubmit}}
```
CvX, line 44:
```suggestion
      @onSubmit={{@onSubmit}}
```

CvX (approved):

assets/javascripts/discourse/components/topic-assignments.gjs
jjaffeux: ah yes shit, we don't have linting here 🙃

PR #568, opened 2024-05-02
DEV: remove assign_path and assign_icon from seralizers

lis2:
Both fields are unnecessary and should not be included in the serializer to reduce payload.


jjaffeux (approved):

jjaffeux:
Do we have existing specs ensuring that the front is still working?

PR #571, opened 2024-05-08
FEATURE: Add pending assign reminder threshold setting

Grubba27:
User can define a threshold for the pending assign reminder.

### DIFF
--- a/lib/pending_assigns_reminder.rb
+++ b/lib/pending_assigns_reminder.rb
@@ -4,7 +4,7 @@ class PendingAssignsReminder
   REMINDED_AT = "last_reminded_at"
   REMINDERS_FREQUENCY = "remind_assigns_frequency"
   CUSTOM_FIELD_NAME = "assigns_reminder"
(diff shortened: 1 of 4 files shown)


lib/pending_assigns_reminder.rb
```diff
@@ -4,7 +4,7 @@ class PendingAssignsReminder
   REMINDED_AT = "last_reminded_at"
   REMINDERS_FREQUENCY = "remind_assigns_frequency"
   CUSTOM_FIELD_NAME = "assigns_reminder"
-  REMINDER_THRESHOLD = 2
+  REMINDER_THRESHOLD = SiteSetting.pending_assign_reminder_threshold
```
markvanlan, line 7: I don't think you can set the const here, as you cannot change the site setting after the server spin up. I think removing the const, and calling `SiteSetting` directly where the const is  currently used should work.

markvanlan (approved):

PR #592, opened 2024-09-09
FIX: If a group is deleted also remove assignments

oblakeerickson:
This fix addresses an issue if a group is deleted topics will remain
assigned to all group members with no way to un-assign those users from
the original group assignment.

Now if a group is deleted any group assignments are also deleted.

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -873,9 +873,11 @@ after_initialize do
   end
 
   on(:group_destroyed) do |group, user_ids|
-    User.where(id: user_ids).find_each do |user|
-      user.notifications.for_assignment(group.assignments.select(:id)).destroy_all if user
-    end
+    User
+      .where(id: user_ids)
+      .find_each do |user|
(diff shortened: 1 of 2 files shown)


plugin.rb
```diff
@@ -872,6 +872,15 @@ module ::DiscourseAssign
     end
   end
 
+  on(:group_destroyed) do |group, user_ids|
+    user_ids.each do |user_id|
+      user = User.find(user_id)
```
lis2, line 877:
```suggestion
      user = User.find_by(id: user_id)
```
Later you have `if user` so we don't want error when a user does not exist

lis2 (approved):

plugin.rb
martin-brennan, line 877:
```suggestion
    User.where(id: user_ids).find_each do |user|
```

Seems much more efficient to find all the users at once, and in batches in case the group has tons of users.

PR #601, opened 2024-11-05
FEATURE: better notification level on assign

ZogStriP:
Use the "when I post in a topic, set that topic to..." user preference as the default notification level when a topic/post is assigned to a user instead of always setting it to "watching".


oblakeerickson (approved):

ZogStriP:
Closing this one as I'll be adding a user preference instead.

PR #602, opened 2024-11-11
DEV: ensure bulk action is done before checking for result

jjaffeux:
Sometimes the target topic can be out of the viewport and this was causing failing specs.

### DIFF
--- a/spec/system/bulk_assign_spec.rb
+++ b/spec/system/bulk_assign_spec.rb
@@ -46,6 +46,7 @@ describe "Assign | Bulk Assign", type: :system do
 
       # Click Confirm
       topic_list_header.click_bulk_topics_confirm
+      expect(assign_modal).to be_closed
 
       # Reload and check that topic is now assigned
       visit "/latest"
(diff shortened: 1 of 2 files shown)


jjaffeux:
Please merge if you approve

jjaffeux:
> Were you able to reproduce the flaky failure?
> 
> According to the capybara docs, the `visible:` parameter relates to element visibility (e.g. things like `display: none`). So theoretically I don't think it help with anything viewport-related. (unless we had viewport-based cloaking/virtual-lists... but our topic-list doesn't have that).
> 
> For viewport-related stuff, capybara has `obscured: true|false`... but that defaults to `nil`, so the viewport shouldn't affect anything. 🤔

I had a lot of different failures, but regardless what you say makes sense... Looking at the latest failure on CI:

image


Clearly it should be unassigned here, but it's not. Given we know it works most of the times, it's probably a timing issue I think we can try to get  away with a `try_until_success`

jjaffeux:
nah... try_until_success it not going to make any difference here

jjaffeux:
@davidtaylorhq ok I think I know what is going:

- start bulk operation
- visit("/latest")

When we visit latest sometimes it probably cancels the request given it happens so fast. If we ask to wait for modal to be closed before we visit latest, that should always be good.

This is why we could have failure on line 52 or 72 as these are the two lines after visit("/latest")

davidtaylorhq (approved):
🚀

PR #604, opened 2024-11-13
FEATURE: Add `should_notify` option to `Assigner#assign`

Grubba27:
This option let's you control whether the added user within a group assignment should be notified or not.

requires https://github.com/discourse/discourse/pull/29741 to ci go green

### DIFF
--- a/app/controllers/discourse_assign/assign_controller.rb
+++ b/app/controllers/discourse_assign/assign_controller.rb
@@ -44,12 +44,7 @@ module DiscourseAssign
       note = params.permit(:note)["note"].presence
       status = params.permit(:status)["status"].presence
       should_notify = params.permit(:should_notify)["should_notify"]
-      should_notify =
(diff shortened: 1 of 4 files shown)


app/controllers/discourse_assign/assign_controller.rb
```diff
@@ -43,6 +43,8 @@ def assign
       group_name = params.permit(:group_name)["group_name"]
       note = params.permit(:note)["note"].presence
       status = params.permit(:status)["status"].presence
+      should_notify = params.permit(:should_notify)["should_notify"]
+      should_notify = (should_notify.present? ? should_notify == "true" : true)
```
davidtaylorhq, line 47: As in the core PR
```suggestion
      should_notify = (should_notify.present? ? should_notify.to_s == "true" : true)
```

davidtaylorhq (approved):

PR #614, opened 2024-12-02
PERF: N+1 when assignments status are enabled

ZogStriP:
This properly preload assignment statuses on topic lists when they're enabled to prevent N+1.

Internal ref - t/142804


brrusselburg (approved):
🙏

PR #620, opened 2024-12-12
FIX: Not send notifications when it should never notify assignment

Grubba27:
- Added silenced_assignments table and migration to store silenced assignments
- Added tests for silenced assignments

### DIFF
--- a/app/jobs/regular/assign_notification.rb
+++ b/app/jobs/regular/assign_notification.rb
@@ -4,6 +4,7 @@ module Jobs
   class AssignNotification < ::Jobs::Base
     def execute(args)
       raise Discourse::InvalidParameters.new(:assignment_id) if args[:assignment_id].nil?
+      return if SilencedAssignment.exists?(assignment_id: args[:assignment_id])
       Assignment.find(args[:assignment_id]).create_missing_notifications!
     end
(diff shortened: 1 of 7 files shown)


app/jobs/regular/assign_notification.rb
```diff
@@ -4,6 +4,7 @@ module Jobs
   class AssignNotification < ::Jobs::Base
     def execute(args)
       raise Discourse::InvalidParameters.new(:assignment_id) if args[:assignment_id].nil?
+      return if SilencedAssignment.exists?(assignment_id: args[:assignment_id])
```
markvanlan, line 7: The most minor style thing -- empty line under early return (guard clauses) is kinda nice.

markvanlan (approved):
In PR / commit description can you write a bit more about the conditions in which this was helpful? It's not obvious how this notification is triggered, especially b/c in `lib/assigner` we see a `should_notify` variable so you'd expect this to already be taken care of.

Looks great! I would normally say we should clean these up.. Which I guess in the assign notification job you could find, delete, and skip rather than find and skip 🤷 . I'm not too worried about this table ballooning though hehe. And you have an index for lookup.

PR #621, opened 2024-12-12
DEV: make specs regarding assign button more lenient

ZogStriP:
in preparation of https://github.com/discourse/discourse/pull/30242 which might make the assign button appear immediately (when it's the only available option) or behind the "..." dropdown.

### DIFF
--- a/test/javascripts/acceptance/assign-disabled-test.js
+++ b/test/javascripts/acceptance/assign-disabled-test.js
@@ -1,4 +1,4 @@
-import { click, visit } from "@ember/test-helpers";
+import { visit } from "@ember/test-helpers";
 import { test } from "qunit";
 import { acceptance } from "discourse/tests/helpers/qunit-helpers";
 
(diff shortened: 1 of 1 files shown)


CvX:
1. Needs a linting fix
2. Is it just in case core adds more buttons and then that turns into a dropdown? 🤔

ZogStriP:
> * Needs a linting fix

Done 👍 



> 2\. Is it just in case core adds more buttons and then that turns into a dropdown? 🤔

Yes, or that the specs are run with another plugins installed

CvX:
Is there actual cross-contamination between plugin tests? are things enabled from other plugins/themes when running d-assign tests?

ZogStriP:
> Is there actual cross-contamination between plugin tests? are things enabled from other plugins/themes when running d-assign tests?

Not in CI (AFAIK), but locally it might.

CvX:
Locally shouldn't be an issue either - unless a test explicitly enables another plugin, its functionality should not be initialized.

CvX (approved):
> that makes things even easier then 😅

love it 😁

PR #626, opened 2025-01-16
FEATURE: new "notification level when assigned" user preference

ZogStriP:
This adds a new user preference allowed users to control how the notification level of the topic they're assigned to changes.

image

Internal ref - t/141162

Dependent on https://github.com/discourse/discourse/pull/30821 being merged.


ZogStriP:
Note: the system spec will keep failing until https://github.com/discourse/discourse/pull/30821 is merged.

lib/assigner.rb
```diff
@@ -506,6 +508,7 @@ def add_small_action_post(action_code, assign_to, text)
       @assigned_by,
       text,
       bump: false,
+      auto_track: false,
```
ZogStriP, line 511: We need this to prevent the "small action" / "whisper" post that is automatically created when assigning a topic from changing the notification level because the user "replied".

assets/javascripts/discourse/initializers/extend-for-assigns.js
```diff
@@ -860,6 +860,8 @@ export default {
 
       api.addUserSearchOption("assignableGroups");
 
+      api.addSaveableUserOptionField("notification_level_when_assigned");
```
ZogStriP, line 863: We need to "globally" allow this user option to be "saveable" when save a user preference's page.

db/migrate/20250115130542_add_notification_level_when_assigned_user_option.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+class AddNotificationLevelWhenAssignedUserOption < ActiveRecord::Migration[7.2]
+  def change
+    add_column :user_options, :notification_level_when_assigned, :integer, null: false, default: 3 # watch topic
```
ZogStriP, line 4: Keep the default of "watching" a topic on assign.

config/locales/client.en.yml
```diff
@@ -100,6 +100,11 @@ en:
               assignable_levels:
                 title: "Who can assign this group"
     user:
+      notification_level_when_assigned: 
+        label: "When assigned"
+        watch_topic: "watch topic"
+        track_topic: "track topic"
+        do_nothing: "do nothing"
```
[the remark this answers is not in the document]
ZogStriP: It's technically the "continuation" of a sentence. "When assigned... watch topic".

PR #634, opened 2025-02-18
FIX: skip notifications when topic is not found

dbattersby:
Fixes a bug where the topic may already be deleted before the notification job runs, causing the job to fail.


Drenmi (approved):
🚀

PR #643, opened 2025-03-14
DEV: Remove code that is not necessary

tgxworld:
Calling `expand_if_needed` is not necessary here.


tgxworld:
Thanks @tyb-talks

PR #648, opened 2025-03-18
FEATURE: add "assigned:<name>" filter (in the /filter page)

ZogStriP:
The "assigned:\<name\>" filter was already available in the advanced search but it wasn't in the /filter page.

This commit adds that filter allowing anyone to filter topics assigned to either a specific user (using their username) or to a specific group (using its name).

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -888,11 +888,11 @@ after_initialize do
   end
 
   add_filter_custom_filter("assigned") do |scope, filter_values, guardian|
-    return if !guardian.can_assign? || filter_values.blank?
+    next if !guardian.can_assign? || filter_values.blank?
 
     user_or_group_name = filter_values.compact.first
 
-    return if user_or_group_name.blank?
+    next if user_or_group_name.blank?
 
     if user_id = User.find_by_username(user_or_group_name)&.id
       scope.where(<<~SQL, user_id)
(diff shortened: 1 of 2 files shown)


plugin.rb
```diff
@@ -887,37 +887,47 @@ module ::DiscourseAssign
     Assignment.active_for_group(group).destroy_all
   end
 
-  register_search_advanced_filter(/in:assigned/) do |posts|
-    posts.where(<<~SQL) if @guardian.can_assign?
-        topics.id IN (
-          SELECT a.topic_id FROM assignments a WHERE a.active
-        )
+  add_filter_custom_filter("assigned") do |scope, filter_values|
+    user_or_group_name = filter_values.compact.first
```
pmusaraj, line 891: Should we gate this also to users with the `can_assign` permission?

spec/requests/list_controller_spec.rb
```diff
@@ -355,4 +355,42 @@
       expect(JSON.parse(response.body)["topic_list"]["topics"].map { |t| t["id"] }).to eq([pm.id])
     end
   end
+
+  describe "#filter" do
+    include_context "with group that is allowed to assign"
```
pmusaraj, line 360: Would be good to have a test case for users that aren't allowed to assign.

pmusaraj (approved):

ZogStriP:
The fixes for the plugins are in https://github.com/discourse/discourse-ai/pull/1206 and https://github.com/discourse/discourse-topic-voting/pull/237

PR #659, opened 2025-05-08
DEV: [gjs-codemod] Convert all templates to gjs

CvX:


assets/javascripts/discourse/connectors/user-messages-nav/assigned-messages.hbs
CvX, line 1: this outlet was removed in https://github.com/discourse/discourse/pull/19294

davidtaylorhq (approved):

